PR #23, opened 2014-04-04
Change file_exists to is_readable

swalkinshaw:
Using `is_readable` is a more comprehensive check since it doesn't really matter if the file exists if it's not also readable. Modified the error message to reflect the other possibility as well.

### DIFF
--- a/src/Dotenv.php
+++ b/src/Dotenv.php
@@ -16,10 +16,10 @@ class Dotenv
         }
 
         $filePath = rtrim($path, '/') . '/' . $file;
-        if (!file_exists($filePath) || !is_file($filePath)) {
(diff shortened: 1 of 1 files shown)


kolesar-andras:
`is_readable()` returns false in some environments, see these PHP tickets:

* [is_readable() returns false on readable files on windows network drive to linux](https://bugs.php.net/bug.php?id=53214)
* [is_readable returns false for Windows Offline File, but file can be read](https://bugs.php.net/bug.php?id=69834)
* [Network drive file: is_readable() == false but file can be read](https://bugs.php.net/bug.php?id=73543)

I had to remove `is_readable()` call to make that environment working.

PR #56, opened 2015-01-02
Reorganize fixtures

lucasmichot:


vlucas:
The instance variable is a much nicer way to do this. Thanks.

PR #58, opened 2015-01-02
Ensure some files are not exported

lucasmichot:


GrahamCampbell:
We should add the tests in here too.

lucasmichot:
@vlucas can I add some more files ?

vlucas:
Sure - just push to this branch agin

.gitattributes
```diff
@@ -0,0 +1,4 @@
+.gitattributes export-ignore
+.gitignore export-ignore
+.travis.yml export-ignore
+tests/ export-ignore
```
GrahamCampbell: shouldn't be a slash?

PR #60, opened 2015-01-02
Add PHP-CS-Fixer level2 for Travis

lucasmichot:


GrahamCampbell:
:-1: from me. I hate the idea of running cs fixers on travis.

.travis.yml
```diff
@@ -10,6 +10,11 @@ php:
 
 sudo: false
 
-install: travis_retry composer install --no-interaction --prefer-source
+install:
+  - wget http://get.sensiolabs.org/php-cs-fixer.phar
```
GrahamCampbell: why are you doing this and requiring it through composer?

lucasmichot:
Why @GrahamCampbell ?
Some other packages also do it : https://github.com/thephpleague/omnipay-common

GrahamCampbell:
Travis should not fail because someone screwed up the cs. We need something separate for that.

vlucas:
I actually don't think this is a terrible idea. It would prevent people from submitting code that does not conform to PSR-2, which is the goal of this. What else would you suggest for this, @GrahamCampbell ?

GrahamCampbell:
@vlucas I still run cs checks on pull requests automatically, just on on travis. Symfony use fabbot.io, I use styleci. https://github.com/StyleCI.

PR #66, opened 2015-02-05
Parse line breaks

b3njamin:
Converts escaped line breaks and makes them literal line breaks. Useful things like adding public or private keys to environment variables.

### DIFF
--- a/src/Dotenv.php
+++ b/src/Dotenv.php
@@ -130,6 +130,7 @@ class Dotenv
     {
         list($name, $value) = static::splitCompoundStringIntoParts($name, $value);
         $name  = static::sanitiseVariableName($name);
+        $value = static::parseLineBreaks($value);
         $value = static::sanitiseVariableValue($value);
         $value = static::resolveNestedVariables($value);
 
@@ -221,6 +222,20 @@ class Dotenv
             );
         }
 
+        return $value;
+    }
+
+    /**
+     * Look for escaped line breaks \n and make them literal line breaks
+     *
+     * @param $value
+     * @return mixed
+     */
+    protected static function parseLineBreaks($value)
+    {
+        if (strpos($value, '\n') !== false || strpos($value, '\r') !== false) {
+            $value = str_replace(array('\n', '\r'), array("\n", "\r"), $value);
+        }
         return $value;
     }
 
--- a/tests/Dotenv/Dotenv.php
+++ b/tests/Dotenv/Dotenv.php
@@ -153,4 +153,26 @@ class DotenvTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals('22222:22#2^{', getenv('SPVAR4'));
         $this->assertEquals("test some escaped characters like a quote \\' or maybe a backslash \\\\", getenv('SPVAR5'));
     }
+
+    public function testDotenvParsesLineBreaks()
+    {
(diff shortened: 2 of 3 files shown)


kelvinj:
:-1: from me - I would dispute this expected behaviour.

b3njamin:
Both node dotenv and ruby dotenv support this.

kelvinj:
For me, that's getting too opinionated for a utility library. 

What happens if you have a password like `s3!kk\n12s8`? Then you'd have to get untangled.

I just tested putting this in a `.env` file and running `source .env` … the newlines were not interpreted.

b3njamin:
How ever if you If you source a `.env` file containing a var in quotes e.g. `export NEW_LINE="this has a new \n line"` & `echo $NEW_LINE` the new lines are interpreted.

GrahamCampbell:
:-1: from me too.

b3njamin:
Heroku also supports multiline env vars.

vlucas:
Shell does support this behavior, as noted by @b3njamin:

```
export NEW_LINE="this has a new \n line" && echo $NEW_LINE
```

Seems like if shell supports it, phpdotenv should too.

kelvinj:
The support for that is coming from `echo` not from the environment.

b3njamin:
Bash interprates literal characters when enclosed in double quotes. 

https://www.gnu.org/software/bash/manual/html_node/Double-Quotes.html

kelvinj:
@b3njamin I can't make head nor tail of that article :smile_cat: In any case, I'm sure this PR doesn't yet cover the nuances outlined.

> Shell does support this behavior, as noted by @b3njamin:
> `export NEW_LINE="this has a new \n line" && echo $NEW_LINE`
> Seems like if shell supports it, phpdotenv should too.

I've tried on 3 different OSes and I don't get that behaviour:

```bash
-bash-4.1$ cat > .envtest
export LINEB1='awesome\nmultiline\nvalue'           
export LINEB2="awesome\nmultiline\nvalue"           
export LINEB3=awesome\nmultiline\nvalue             
export LINEBVAR=crazy                               
export LINEB4=awesome\n{$LINEBVAR}\nmultiline\nvalue
^C
-bash-4.1$ source .envtest 
-bash-4.1$ echo $LINEB1
awesome\nmultiline\nvalue
-bash-4.1$ echo $LINEB2
awesome\nmultiline\nvalue
-bash-4.1$ echo $LINEB3
awesomenmultilinenvalue
-bash-4.1$ echo $LINEB4
awesomen{crazy}nmultilinenvalue
-bash-4.1$
```

boynux:
@kelvinj you have to pass `-e` to echo to enable interpretation of back slashes. If you try `echo $LINEB2` in above snippet you should get something like:

    awsome
    multiline
    value

boynux:
I guess the best way to support this particular feature is to enable different interpretation for `double quote` and `single quote` (as in Ruby Dotenv).  
If you need to keep `\n` as literal then you should use `single quote` and `double quote` otherwise. 
It's natural to use, since PHP behaves the same way when dealing with escape characters in strings.

freezy-sk:
@boynux :+1:

GrahamCampbell:
Can this be closed now?

GrahamCampbell:
In V3, we now support multiline strings properly.

PR #67, opened 2015-02-19
Also generate SYMFONY__* env vars

mattallty:
Allow Symfony to use generated env variables by prefixing them with "SYMFONY__". 
See [Symfony docs](http://symfony.com/doc/current/cookbook/configuration/external_parameters.html).

### DIFF
--- a/src/Dotenv.php
+++ b/src/Dotenv.php
@@ -73,8 +73,9 @@ class Dotenv
         }
 
         putenv("$name=$value");
-        $_ENV[$name] = $value;
-        $_SERVER[$name] = $value;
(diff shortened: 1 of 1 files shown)


vlucas:
This should be an optional configuration or method call. This change will pollute everyone's environment variables for the benefit of only Symfony users.

GrahamCampbell:
:-1: on this. Symfony should wrap this itself if it needs this hacky crap.

kelvinj:
:-1: I agree with Graham.

gonimar:
:-1:

PR #68, opened 2015-02-19
Add function to expose current status of the $immutable variable

Ultrabenosaurus:
There are functions to change between mutable and immutable, but due to the `$immutable` variable being a `protected` member it is difficult to get the current status.

I have added a function that simply returns the `$immutable` variable, as well as tests to ensure it works as expected.

### DIFF
--- a/src/Dotenv.php
+++ b/src/Dotenv.php
@@ -244,6 +244,15 @@ class Dotenv
     }
 
     /**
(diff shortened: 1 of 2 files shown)


vlucas:
Thanks! :+1:

Ultrabenosaurus:
No worries :)

PR #73, opened 2015-03-08
In PHP value "0" is false

boynux:
In PHP if `$value = "0"` then stetmentes like `if (!$value)` will be `true`, which might not be an expected behavior in this case.
If you check Ruby Dotenv code, same statement works pretty well and that's because in Ruby value `0` is equivalent to `true` (unlike most other languages).

### DIFF
--- a/src/Dotenv.php
+++ b/src/Dotenv.php
@@ -161,7 +161,7 @@ class Dotenv
     protected static function sanitiseVariableValue($value)
     {
         $value = trim($value);
-        if ($value === null) return '';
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Your change broke the tests.

boynux:
Haha,
Correct, I didn't run tests. Can you fix test please.
You got the bug I mentioned right?
I can do it as well but might take some time since I don't have much time
during working days.

vlucas:
This is fixed in v2.0.0.

PR #79, opened 2015-04-22
prefix json fixture variables with J

xjchengo:
The json fixture variable names is same with php fixture variable names. After DotenvJsonTest, the variables have been set and when loading, the default $immutable is true. So even if php fixtures don't be loaded, the test can pass.


vlucas:
Good idea. :+1:

PR #81, opened 2015-04-23
make nested variable escapable

xjchengo:


xjchengo:
Sorry, the unittest didn't pass. I will fix it.

PR #82, opened 2015-04-23
Nested variable escapable

xjchengo:
Because of sanitiseVariableValue method in EnvLoader, `$` could not be escaped by single backslash. Maybe we need to rewrite sanitiseVariableValue to support single backslash escape.


vlucas:
@xjchengo we can still support this in the v2.x release - please apply all these changes to the new v2.0.0 release and submit a new PR and we'll take a look.

PR #87, opened 2015-05-29
Initial v2.0 version (for really reals this time)

vlucas:
Changes made:

 * Moved everything under a proper PSR-4 `Dotenv` namespace with a constructor, so usage is now:
```php
$dotenv = new Dotenv\Dotenv(__DIR__);
$dotenv->load();
```
 * Now loads immutable by default (will not overwrite existing ENV vars). You can use `$dotenv->overload()` to make them mutable
 * Now supports only shell-style variable syntax, and no other alternate loaders
 * Updated README file to match new sytnax, etc.
 * Kept same `required` assertions and others from the original `v2` branch


README.md
```diff
@@ -186,6 +211,16 @@ This can be achieved via an automated deployment process with tools like
...
+
+If you need to use environment variables that you have set in your `.env` file
+in a command line script that doesn't use the Dotenv library, you can `source`
+it into your local shell session:
+
+```
+source .env
+```
+
 Contributing
```
sagikazarmark: Shouldn't headline markup be consistent?

vlucas: It is consistent with the other headlines in the file. You can't do an h3 level heading without the hash in Markdown.

PR #96, opened 2015-06-01
add exception classes

xjchengo:
All exceptions extend from InvalidArgumentException. So it is backwards-compatible at most time. But if anyone relies on exception message, it maybe failed. 
Or we put same exception message in FileNotFoundException and FilePermissionException.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -232,7 +232,7 @@ class Loader
 
             // Unquoted values cannot contain whitespace
             if (preg_match('/\s+/', $value) > 0) {
-                throw new SyntaxException();
+                throw new SyntaxException('Dotenv values containing spaces must be surrounded by quotes.');
             }
         }
 
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -69,8 +69,8 @@ class Loader
     /**
      * Ensures the given filePath is readable.
      *
-     * @throws FileNotFoundException
-     * @throws FilePermissionException
+     * @throws \Dotenv\Exception\FileNotFoundException
+     * @throws \Dotenv\Exception\FilePermissionException
      *
      * @return void
      */
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -34,6 +34,8 @@ class Loader
      *
      * @param string $filePath
      * @param bool   $immutable
+     *
+     * @return void
      */
     public function __construct($filePath, $immutable = false)
     {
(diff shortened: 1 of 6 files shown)


src/Loader.php
```diff
@@ -32,8 +34,6 @@ class Loader
      *
      * @param string $filePath
      * @param bool   $immutable
-     *
-     * @return void
```
GrahamCampbell: Keep this please.

src/Loader.php
```diff
@@ -67,17 +67,27 @@ public function load()
     /**
      * Ensures the given filePath is readable.
      *
-     * @throws \InvalidArgumentException
+     * @throws FileNotFoundException
```
GrahamCampbell: Please use the FQCN is all phpdoc.

GrahamCampbell:
Could you refactor so that each exception class generates the message instead of having it passed to each exception.

xjchengo:
Yes. I am thinking about backwards-compatible. See my above words please.

GrahamCampbell:
My suggestion wouldn't be breaking anything?

GrahamCampbell:
The behaviour would remain the same as it is in this PR, only each exception is responsible for it's own message.

xjchengo:
Your suggestion is fine. It is about my implementation. My current implementation has changed exception message. I worried about someone relying on exception message.

GrahamCampbell:
> I worried about someone relying on exception message.

Nah, it's fine to change it.

GrahamCampbell:
We can always bump the major version on the master branch again/

xjchengo:
Ok, I got it.

GrahamCampbell:
Why have you closed this?

src/Loader.php
```diff
@@ -231,7 +232,7 @@ protected function sanitiseVariableValue($name, $value)
 
             // Unquoted values cannot contain whitespace
             if (preg_match('/\s+/', $value) > 0) {
-                throw new InvalidArgumentException('Dotenv values containing spaces must be surrounded by quotes.');
+                throw new SyntaxException();
```
vlucas: I still want the ability to pass a message here -  A `SyntaxException` could be a variety of errors - not just this one (though it is only this one at this time).

GrahamCampbell: Maybe make the message param optional like i do in factory muffin.

xjchengo: There will be other type syntax exception future. 
```
'Dotenv values containing spaces must be surrounded by quotes.'
```
is not a base exception message. Using it as default is weird. 
How about making error message required parameter of SyntaxException construstor ?

vlucas: Yes - make it a parameter. That is the expected interface for the Exception class. I don't want to deviate from it.

PR #97, opened 2015-06-10
add composer/installers to allow the option to customize where the li…

matgargano:
…brary is loaded

I have a WordPress application where I am going to need to specifically install this library in a separate directory and include it earlier than it would if it were in my default vendor directory.  I'm already using the vendor directory in a separate location (`mu-plugins` with the composer autoloader) which runs much later than I need the dotenv code to run.

This shouldn't impact anything in your app.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,8 @@
         }
     ],
     "require": {
-        "php": ">=5.3.9"
+        "php": ">=5.3.9",
+        "composer/installers": "~1.0"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.0"


GrahamCampbell:
You don't need this.

matgargano:
Care to explain?

vlucas:
I have not used this package before, but I am confused as to why we should use this. The phpdotenv package is not a Wordpress plugin and should not be installed as such for the vast majority of our users. Based on what I have just read about the composer/installer package, this might have adverse effects for everyone else that uses and depends on this package. Am I understanding how this package works correctly?

matgargano:
I am not using this as a "Wordpress" package. I am using this as a generic library in front of my wordpress application.

I could just include my auto loader and be set - except for the fact that I'm using my vendor directory to include libraries in my muplugins directory which loads much later than I need the dotenv package. 

Adding this dependency would basically allow the explicit setting of the directory to put this package. It should not adverse impact any applications using this package. 

I  could always fork this and maintain it but adding in this package will not impact the consumption of this library- Can you provide a use case where it would?

GrahamCampbell:
You don't need to require this package here. Require it in your own code.

matgargano:
```
this might have adverse effects for everyone else that uses and depends on this package.
```

Can you provide a use case where it would?

GrahamCampbell:
I literally have no idea what you're talking about.

matgargano:
What is your reasoning for not accepting this pull request? From the dialog above it appears it's because of what @vlucas said `this might have adverse effects for everyone else that uses and depends on this package.` since @GrahamCampbell has not given a reason outside of vague comments. I was asking for what adverse effect it could have - which was really my way of saying why are you not accepting the PR?

If you don't it's not a big deal, it's more a convenience for me, I can fork it myself and require my fork.

GrahamCampbell:
The reasoning is this PR makes no sense what so ever. Never seen anything quite so silly as a suggestion.

prisis:
@matgargano take a look on https://roots.io/bedrock/ how they use php do tend.

GrahamCampbell:
Doesn't mean they have a clue what they're doing, and also you're wrong. Their actual packages are NOT requiring it. https://github.com/roots/wp-cli-theme-activation/blob/master/composer.json

matgargano:
wat

vlucas:
Okay @GrahamCampbell let's take a step back. I have never used this composer/installer package, and I think the best course of action is to try to understand what it is and how it works before shutting down the conversation entirely. Based on the conversation so far, I don't think either of us know exactly how this package works and/or how it helps people working with WordPress, etc. I for one would like to understand it better so I can make a better judgement.

@matgargano I read the website for this package, and the way I understand it is:

1. Require the `composer/installer` package
2. Specify a `type` that declares what software *your* library is for
3. Optionally specify a specific directory to install your library into

My confusion here is that neither (2) or (3) would apply to phpdotenv, which leads me to believe that we either don't need it, or I am missing something. Can you fill us on on exactly how this works, and why putting this as a requirement in phpdotenv will help *your* packages?

GrahamCampbell:
This installer package does NOT nned this package to require it. That much is pretty obvious. Composer would be pretty shit if that were to be the case.

@vlucas You are of course correct. This has nothing to do with this library.

matgargano:
@vlucas thank you let me collect my thoughts and elucidate my issue. it comes down to composer being a dependency manager ... but in times it acts as an installer...

matgargano:
@vlucas ok so, the way I am using composer in my workflow is, if the package type is "wordpress-plugin" I will have composer install it into `wp-content/plugins` and if the package type is `library` it is installed in `wp-content/mu-plugins`

the `wp-content/plugins` directory works like such:
 - plugins are individual files or directories that have a php file that has specific boilerplate at the top of them indicating they are a plugin
 - these plugins are available in a Plugins page in the WordPress administration area and can be turned on and off (when they are turned on, during the bootstrapping process the main plugin file [the one with the boilerplate]) is included. `wp-content/plugins` is a moot point for this discussion.

 the `wp-content/mu-plugins` directory works like such:
 - any file in the root of that directory is executed, so if you put a folder in that directory without a file in the root to include it, the plugin will never get loaded. Sounds like a perfect case to have composer install dependencies there right?

 The drawback is that code is executed after wordpress connects to the database, so relying on composer's autoloader is out of the question.  So without any explicit requiring done manually (including a file in a subdirectory in the mu-plugins directory, which is risky in case your environment changes, since the mu-plugins directory isn't always in the same location, I've seen workflows where it's in `wp-content/mu-plugins` (standard) and also `app/mu-plugins` among others)  

 That's where `composer/installers` comes into play. Requiring this package allows you to specify where to install specific packages, see https://github.com/composer/installers where it gives you an example: 

 ```
 {
    "extra": {
        "installer-paths": {
            "your/custom/path/{$name}/": ["shama/ftp", "vendor/package"]
        }
    }
}
```

So i can specify where I want phpdotenv to live so that i can have a predictable location to include.

Since the package type is "library" this would only have an issue if somebody is specifying where a composer package `library` would go, which is, effectively the same as the vendor directory. 

This shouldn't cause a problem, but it may be a philosophical issue: should composer be an autoloader + package manager or give some installation functionality, which it seems like it's in limbo as far as its role is. I know it's stated role is not an installer but it's built to act as one.

**Comment by tuupola** [comment]
I fail to understand why `composer/installers` should be in phpdotenv `composer.json` file. To me it sounds like it should be required in the plugins `composer.json` file. It is the plugin using features provided by  `composer/installers` and not phpdotenv.

**Comment by prisis** [comment]
Composer is not a package manager. Yes, it deals with "packages" or libraries, but it manages them on a per-project basis, installing them in a directory (e.g. vendor) inside your project. By default it will never install anything globally. Thus, it is a dependency manager.

So set you composer like this [composer.json](https://github.com/gwa/bedrock-multisite-skeleton/blob/master/composer.json) just put phpdotenv not in dev dependency [require-dev] but in dependency [require]

add ``require_once(dirname(__DIR__) . '/vendor/autoload.php');``` into wp-config.php

**Comment by matgargano** [comment]
@prisis my vendor directory *is* the mu-plugins directory, so this does not work.

**Comment by matgargano** [comment]
Solved this problem using  `mnsami/composer-custom-directory-installer`

This conversation is over from my end.

**Comment by vlucas** [comment]
I am glad to hear you got this fixed without any changes to phpdotenv, and sorry that the conversation was not more friendly from the outset. :+1:

**Comment by matgargano** [comment]
Me as well sir. Thanks for taking over as a voice of reason.

**Comment by bristweb** [comment]
I know this has been a while, but thanks for posting [the solution](#issuecomment-111277100) matgargano. 

In case it is helpful to anyone else, one other possibility would be to set custom directories for your mu-plugins with ```wordpress-muplugin``` according to the [docs](https://github.com/composer/installers/blob/master/README.md)

And then move the main vendor directory wherever you want composer.json ex:
```
  "config": {
    "vendor-dir": "some/sub/folder/vendors"
  },
```

FWIW, the combativeness in this thread was exhausting to read.  Adding ```composer/installers``` sounds like a good idea to me, but apparently not a common request.

matgargano:
You got it @bristweb

GrahamCampbell:
It makes zero sense to add this on any level.

PR #102, opened 2015-07-13
Move loader creation to the constructor

### LINKED ISSUE #101: Validation is impossible in 2.0 release without explicitly loading .env from disk
Previous version allowed to perform validation without having to load `.env` file from disk, which is very useful for production set-up, e.g. as discussed [here](https://github.com/vlucas/phpdotenv/issues/63#issuecomment-89490588):

```php
// Load required environment variables from .env in development
$dotenv = new Dotenv\Dotenv();
if(BULLET_ENV == 'development') {
    $dotenv->load(dirname(__DIR__), 'dev.env');
}
// Required ENV vars for this app in production
$dotenv->required([
    'DATABASE_URL',
    'PATH_FILE_UPLOADS',
    'GOOGLE_PLACES_API_KEY',
    'RAVEN_DSN',
    'MANDRILL_USER', 'MANDRILL_PASS'
]);
```

However if you try to do the same with latest release (directory should now be specified in `Dotenv` constructor rather than `load`method, but this is minor), PHP will throw an exception because it tries to initialize `Validator` by passing in `loader` which is `null`. That practically makes it impossible to use validation in production.. why was it done? How can I do validation without loading `.env`?
```
PHP Catchable fatal error:  Argument 2 passed to Dotenv\\Validator::__construct() must be an instance of Dotenv\\Loader, null given, called in /var/www/example.com/vendor/vlucas/phpdotenv/src/Dotenv.php on line 83 and defined in /var/www/example.com/vendor/vlucas/phpdotenv/src/Validator.php on line 34
```

louim:
This allow the use of `$dotenv->required` to validate environment variables without loading without explicitly loading a .env file. Unless i'm horribly mistaken, this should fix #101 and allow validation environment variables even if we don't use a .env file.

### DIFF
--- a/src/Dotenv.php
+++ b/src/Dotenv.php
@@ -26,6 +26,7 @@ class Dotenv
     public function __construct($path, $file = '.env')
     {
         $this->filePath = $this->getFilePath($path, $file);
+        $this->loader = new Loader($this->filePath, $immutable = true);
     }
 
     /**
@@ -35,8 +36,6 @@ class Dotenv
      */
     public function load()
     {
-        $this->loader = new Loader($this->filePath, $immutable = true);
-
         return $this->loader->load();
     }
 
--- a/tests/Dotenv/DotenvTest.php
+++ b/tests/Dotenv/DotenvTest.php
@@ -271,4 +271,14 @@ class DotenvTest extends \PHPUnit_Framework_TestCase
         $dotenv->load();
         $dotenv->required('ASSERTVAR3')->notEmpty();
     }
+
+    /**
+     * @expectedException RuntimeException
+     * @expectedExceptionMessage One or more environment variables failed assertions: foo is missing
+     */
+    public function testDotenvValidateRequiredWithoutLoading()
+    {
+        $dotenv = new Dotenv($this->fixturesFolder, 'assertions.env');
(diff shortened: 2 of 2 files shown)


GrahamCampbell:
Could you include a test case that failed before this change please?

louim:
@GrahamCampbell Done.

GrahamCampbell:
:beers:

src/Dotenv.php
```diff
@@ -35,8 +36,6 @@ public function __construct($path, $file = '.env')
      */
     public function load()
     {
-        $this->loader = new Loader($this->filePath, $immutable = true);
```
vlucas: I don't think this can be removed, but there is not a test case to cover the scenario for why.

Basically, if you call `overload`, and then `load` on a different file, the same loader will be retained from `overload`, which will leave the `immutable` variable set to `false` (from the `overload` method), and the expected behavior will diverge from what this method is supposed to do. A test case needs to be added to ensure this does not happen, and this line probably needs to be added back in. @GrahamCampbell

louim: So, if i'm getting this right, You would have the line in the constructor and in the `load` method, and the load method would overwrite the "default" loader with a new one?

Issue #106, opened 2015-07-26
Ability to clear an environment variable


thedavidmeister:
It's nice to be able to set and get environment variables from `.env` when developing generally, but also nice to set and get *specific* environment variables when writing tests for functionality that behaves differently depending on config.

To write a test that sets an environment variable, then checks some "thing", then clears the environment variable, ready for the next test that just uses the Dotenv provided defaults, we have to do this in the first test as part of `tearDown()`:

````
setenv('FOO');
unset($_ENV['FOO']);
unset($_SERVER['FOO']);
````

Not the end of the world, but maybe Dotenv could provide this method on the Loader as `clearEnvironmentVariable` (respecting the immutability-ability of it all).

thedavidmeister:
I've put a PR up, with tests.

PR #107, opened 2015-07-26
Clear variables

thedavidmeister:
This is for issue #106


vlucas:
Thanks for the code - I always like PRs with tests.

PR #110, opened 2015-07-30
Fix nested section in documentation

martinbean:


vlucas:
Are you using v1 or v2? Because the existing syntax was correct for v2, and your edits are the syntax for v1.

martinbean:
@vlucas I’m not sure. I’m using the version that ships with Laravel 5.1.*.

jessedc:
I cam here with exactly the same issue as @martinbean, never thought to look at changes in phpdotenv versions, or that laravel might be using an older version.

[Laravel Framework 5.1 is still using phpdotenv ~1.0](https://github.com/laravel/framework/blob/5.1/composer.json#L42).

It might be worth amending this PR to mention the version 1 syntax along side version 2 - I imagine a lot of developers discovering this project from lumen/laravel.

GrahamCampbell:
> @vlucas I’m not sure. I’m using the version that ships with Laravel 5.1.*.

Laravel 5.1 uses phpdotenv 1, Laravel 5.2 uses phpdotenv 2.

PR #114, opened 2015-08-17
Enhancement: Use SVG badge for displaying Travis build status

localheinz:
This PR

* [x] uses an SVG badge for displaying the Travis build status, as it is more pleasing for the eye


vlucas:
Wow, the font on the SVG badge looks *way* better. :+1:

PR #116, opened 2015-08-17
Fix: Assert that InvalidArgumentException is thrown

localheinz:
This PR

* [x] asserts that an `InvalidArgumentException` is thrown when `Dotenv` failed to load a file


vlucas:
Thanks! :+1:

localheinz:
You're welcome, @vlucas!

PR #117, opened 2015-08-18
Validations

BruceDev:
Added an integer validation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,21 @@ Beyond simply requiring a variable to be set, you might also need to ensure the
 variable is not empty:
 
 ```php
+$dotenv->required('FOO')->isInteger();
+```
+
+If the environment variable is not an integer, you'd get an Exception:
+
+```
+One or more environment variables failed assertions: FOO is not an integer
+```
+
+### Empty Variables
+
+Beyond simply requiring a variable to be set, you might also need to ensure the
+variable is not empty:
+
+```php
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -168,6 +168,21 @@ If the environment variable is empty, you'd get an Exception:
 One or more environment variables failed assertions: DATABASE_DSN is empty
 ```
 
+### Integer Variables
+
+Beyond simply requiring a variable to be set, you might also need to ensure the
+variable is not empty:
```
vlucas: Please update the description to match your integer validation :)

src/Validator.php
```diff
@@ -60,6 +60,21 @@ function ($value) {
...
     /**
+     * Assert that each variable is a number.
+     *
+     * @return \Dotenv\Validator
+     */
+    public function isInteger()
+    {
+        return $this->assertCallback(
+            function ($value) {
+                return (is_int($value) == true);
```
vlucas: I am wondering if this will ever actually be true. ENV variables are always strings, so I don't think this will ever pass. What you need to do instead is to use something like `ctype_digit` or a regex pattern like `[0-9]+` to test the actual contents of the string instead of the PHP type.

vlucas:
Also this needs a test - just make a new ENV variable in one of the fixtures files that sets an integer, and then test that the rule passes on it. You will also need a second test to ensure the validation rule fails on a variable without all numeric characters.

PR #140, opened 2015-11-18
No need to tests for PHP < 5.3.9

lucasmichot:
See https://github.com/vlucas/phpdotenv/blob/master/composer.json#L16


GrahamCampbell:
There is no such php version 5.3.9 on travis.

PR #143, opened 2015-12-02
refactored environment and variable to own objects

jmatosp:
This is a work in progress.

We should change environment to "persist" the values only after the validation has run.
Loader now is just a file reader, so probably best with a SplFileObject.
Lots of tests missing, but easy writeable.

This PR works with current tests and no code change required.

Please comment if I should continue.


src/Dotenv.php
```diff
@@ -36,9 +42,8 @@ public function __construct($path, $file = '.env')
      */
     public function load()
     {
-        $this->loader = new Loader($this->filePath, $immutable = true);
-
-        return $this->loader->load();
+        $this->environment->setImmutable(true);
+        return $this->parseLines();
```
jmatosp: remove setImmutable from here, by default is Immutable

src/Dotenv.php
```diff
@@ -23,10 +23,16 @@ class Dotenv
...
+    /**
+     * @var Environment
+     */
+    private $environment;
+
     public function __construct($path, $file = '.env')
     {
-        $this->filePath = $this->getFilePath($path, $file);
-        $this->loader = new Loader($this->filePath, $immutable = true);
+        $this->filePath     = $this->getFilePath($path, $file);
```
GrahamCampbell: Please don't align this

GrahamCampbell:
This PR is quite large. Is there any chance you could separate it install smaller, non-breaking chunks?

jmatosp:
not easy to split, but will give it a try

PR #144, opened 2015-12-17
Everything crashes if $_ENV is missing

mickrip:


rquadling:
OOI, what is the normal operation to have no $_ENV? Unless ```unset($_ENV);``` of course, but what would do that in the normal course of things. No PATH (at a minimum).

src/Loader.php
```diff
@@ -305,6 +305,7 @@ protected function beginsWithAQuote($value)
      */
     public function getEnvironmentVariable($name)
     {
+        if(!isset($_ENV))$_ENV=array();
```
GrahamCampbell: Please follow PSR-2.

PR #147, opened 2016-01-04
1. new environment class to represent environment and its state/actions

jmatosp:
First step to re factor:

New environment class witch will have Environment state and actions on it
ATM only if its immutable or not.

Next step:
 Move methods related to operations on Environment to this class

### DIFF
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Dotenv;
+
+/**
+ * Represents System Environment and actions on it
+ *
+ */
+class Environment
+{
+    /**
+     * Are we immutable?
+     *
+     * @var bool
+     */
+    protected $immutable = true;
+
+    /**
+     * @param boolean $immutable
+     *
+     * @return void
+     */
+    public function setImmutable($immutable)
+    {
+        $this->immutable = $immutable;
+    }
+
+    /**
+     * @return boolean
+     */
+    public function isImmutable()
+    {
+        return $this->immutable;
+    }
+}
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -22,11 +22,9 @@ class Loader
     protected $filePath;
 
     /**
-     * Are we immutable?
-     *
-     * @var bool
+     * @var Environment
      */
-    protected $immutable;
+    protected $environment;
 
     /**
      * Create a new loader instance.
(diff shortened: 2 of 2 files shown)


src/Environment.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace Dotenv;
+
+/**
+ * Represents System Environment and actions on it
+ *
```
GrahamCampbell: cs

src/Environment.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace Dotenv;
+
+/**
+ * Represents System Environment and actions on it
```
GrahamCampbell: cs

src/Environment.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    /**
+     * Are we immutable?
+     *
+     * @var bool
+     */
+    protected $immutable = true;
+
+    /**
+     * @param boolean $immutable
```
GrahamCampbell: cs

src/Environment.php
```diff
@@ -0,0 +1,35 @@
...
+     *
+     * @return void
+     */
+    public function setImmutable($immutable)
+    {
+        $this->immutable = $immutable;
+    }
+
+    /**
+     * @return boolean
```
GrahamCampbell: cs

src/Loader.php
```diff
@@ -22,11 +22,9 @@ class Loader
     protected $filePath;
 
     /**
-     * Are we immutable?
-     *
-     * @var bool
+     * @var Environment
```
GrahamCampbell: cs

GrahamCampbell:
Im :-1: here tbh. This adds unneeded complexity.

jmatosp:
The opposite, smaller code parts, easier to maintain, and OOP. probably more lines of code

GrahamCampbell:
I think it's silly to add a class to encapsulate one variable. This only makes sense if we had more than one.

GrahamCampbell:
Thanks anyway.

jmatosp:
this was to break down  the previous PR in to smaller chunks, obviously  this class would end up bigger.

GrahamCampbell:
> obviously this class would end up bigger.

What I meant was send minimal changes.

GrahamCampbell:
I don't think anything really needs to be changed though, on reflection.

PR #149, opened 2016-01-09
Adds missing docbloc on Dotenv constructor

MarkVaughn:
just a little nitpick


GrahamCampbell:
Thanks, but this isn't quite right. I'll patch this.

PR #150, opened 2016-01-11
Fix occurrences of CRLF line endings.

rabbitfang:
- Convert CRLF line endings in exception classes to LF.
- Add rule to .gitattributes to commit all files with LF line endings.


GrahamCampbell:
Thanks. :heart:

PR #155, opened 2016-01-21
Removes unnecessary variables.

systemovich:
* Fixes typos.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -2,15 +2,15 @@
 
 namespace Dotenv;
 
-use Dotenv\Exception\InvalidFileException;
-use Dotenv\Exception\InvalidPathException;
+use Dotenv\Exception\InvalidFileException,
+    Dotenv\Exception\InvalidPathException;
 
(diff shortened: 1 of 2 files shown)


src/Loader.php
```diff
@@ -2,15 +2,15 @@
 
 namespace Dotenv;
 
-use Dotenv\Exception\InvalidFileException;
-use Dotenv\Exception\InvalidPathException;
+use Dotenv\Exception\InvalidFileException,
+    Dotenv\Exception\InvalidPathException;
```
GrahamCampbell: No.

GrahamCampbell:
Lots of these changes are incorrect so I cannot accept this PR.

src/Loader.php
systemovich: PSR-2?

systemovich:
@GrahamCampbell, the only other problem I can see is the `@return` in the doc comment of some functions. What else did I miss?

PR #164, opened 2016-02-22
Check if `putenv` exists before use

litp:
Some platforms disable `putenv` function for security reasons. To make phpdotenv function well on those platforms, we need to check if `putenv` exists before calling it.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -340,9 +340,11 @@ class Loader
         if (function_exists('apache_getenv') && function_exists('apache_setenv') && apache_getenv($name)) {
             apache_setenv($name, $value);
         }
-
-        putenv("$name=$value");
-
+        
+        if (function_exists('putenv')){
+            putenv("$name=$value");
+        }
+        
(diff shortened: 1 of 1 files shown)


src/Loader.php
```diff
@@ -334,9 +334,11 @@ public function setEnvironmentVariable($name, $value = null)
         if ($this->immutable && $this->getEnvironmentVariable($name) !== null) {
             return;
         }
-
-        putenv("$name=$value");
-
+        
+        if (function_exists('putenv')){
```
GrahamCampbell: I feel like we should precompute this and store it as a property.

GrahamCampbell:
Please can you rebase this?

GrahamCampbell:
Hmm, no need to pre-compute, like we did for the other fixes.

GrahamCampbell:
Thanks, I've implemented something very similar.

GrahamCampbell:
It'll be in 2.4.0.

PR #166, opened 2016-02-27
Must execute load() before the required()

MLoureiro:
I took a while to understand why the required env was not being considered/loaded.
I actually needed to look at how the code worked before making it work.

I'm suggesting a simple doc change.
However it would be possible to force load() when executing required()

What you guys think?
Did I missed something?

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,8 @@ If any ENV vars are missing, Dotenv will throw a `RuntimeException` like this:
 One or more environment variables failed assertions: DATABASE_DSN is missing
 ```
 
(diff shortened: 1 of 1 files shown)


vlucas:
Actually this is by design. This is so that phpdotenv can be used to verify ENV vars are present without having to load and .env file. For instance, I load the .env file on my local development environment, but not on production. However, I still need to make sure the ENV var was loaded on production, where no .env file exists.

MLoureiro:
Ok, makes sense.
But shouldn't that be explained somehow somewhere ?

vlucas:
Yes, is probably should be in the README somewhere.

PR #169, opened 2016-04-13
Switch to new recommended syntax

vinkla:
Composer now recommends using `||` instead of a single `|`.

https://getcomposer.org/doc/articles/versions.md#range


vlucas:
Thanks. :+1:

PR #172, opened 2016-04-15
Update phpunit config

vinkla:
We can prevent risky tests, warnings and tests that doesn't test anything.


vinkla:
It seems `testDotenvRequiredCanBeUsedWithoutLoadingFile` fails.

Issue #39, opened 2014-10-17
Overriding Apache env variables


levacic:
So some time ago I had a problem where a certain environment variable was configured in Apache's `<VirtualHost>` block, and `phpdotenv` wasn't overwriting that value with what was configured in the `.env` file.

I use `getenv()` calls in my config files instead of `$_ENV` to simplify handling non-configured values - otherwise I'd have to do something like `array_key_exists($_ENV, 'key')` or something, which is too verbose with no benefit.

The problem with that is, when you have, for example `SetEnv APP_ENV production` in Apache's `<VirtualHost>` block, then the code

```php
putenv('APP_ENV=staging');
echo getenv('APP_ENV');
```

will output `production`.

To successfully override this, PHP's `apache_setenv` function must be called instead, like this:

```php
if (function_exists('apache_setenv'))
{
    apache_setenv('APP_ENV', 'staging');
}
```

The `function_exists()` check must be present so as not to cause errors when the user is not running Apache.

Now of course, this depends on a higher-level decision of whether an Apache-configured environment variable should take precedence over the one specified in a project's `.env` file - and to me it's kind of obvious that it shouldn't, since an `.env` file is a more granular configuration method, and as such, it should take precedence, but of course I may be wrong about that.

Do you think this could/should be included in `phpdotenv`?

---

Also, as an unrelated question (and sorry for hijacking this issue, I just want to avoid spamming issues), is there any reason why you chose to use this specific syntax for `.env` files, other than being similar to Ruby's dotenv syntax? Just plain requiring PHP files (as e.g. Laravel < 5 does) seems to be a much more straightforward approach, which avoids the various issues related to actually parsing the `.env` files, including comments etc. and additionally simplifies generating them, as you can simply `var_export()` an array and be done with it - with the current syntax, it doesn't seem to be as trivial, as one must be careful to correctly escape stuff since the syntax isn't really documented completely (e.g. there's nothing in the docs to explain what to do if I want the literal `{$BASE_DIR}/tmp` as a variable value, without resolving `{$BASE_DIR}` to something else). I'm interested in whether you feel there is any advantage to the current approach over simply doing `$variables = require '/path/to/project/.env.php';` - thanks in advance for your thoughts on this, and sorry for the longish-double-issue.

Thanks for your work, cheers!

GabeMedrash:
[...]

PR #176, opened 2016-04-27
if Loader is not immutable, override Apache env vars

GabeMedrash:
This PR fixes https://github.com/vlucas/phpdotenv/issues/39.

I have not written additional tests due to the added complexity of introducing Apache as a dependency of the test suite, but I would be happy to provide some if you would like.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -336,7 +336,9 @@ class Loader
         }
 
         if (function_exists('apache_setenv')) {
-            if (function_exists('apache_getenv') && apache_getenv($name)) apache_setenv($name, $value);
+            if (function_exists('apache_getenv') && apache_getenv($name)) {
+                apache_setenv($name, $value);
+            }
         }
 
         putenv("$name=$value");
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Please follow PSR-2.

GabeMedrash:
Thanks for the review--I ran `phpcs --standard=PSR2 src` and moved the body of the nested `if` within curly braces. The lint now runs clean. Let me know if there's something else you had in mind.

GrahamCampbell:
Nearly there. Please get rid of the nested if too. ;)

GabeMedrash:
Gotcha--think I got it this time!

GrahamCampbell:
Thanks. Looks good to me. Ping @vlucas for approval.

vlucas:
Looks good to me. The only question I have is that we can probably only check if one of the apache get/set functions exist, because if the module is enabled, both of them will always exist. It's more of a minor thing though, and not a reason to hold up this PR.

GabeMedrash:
@vlucas Thanks, that makes a lot of sense. Will modify now.

vlucas:
Thanks 👍

PR #177, opened 2016-04-28
Falsey value for 'false'

Daerdemandt:
As of now, there is a following gotcha: `THIS_IS_FALSE=false ` results in `getenv('THIS_IS_FALSE')` returning literal 'false', which is truthy.

That is confusing! Especially since editors recognise .env syntax and highlight that 'false' as a recognised value.

I suppose parsing lowercase unquoted 'false' as a falsey value is less surprizing than parsing it as a truthy one.

Alternatively, this gotcha could be mentioned in the Readme with a suggestion to use '1' and '0' instead of 'true' and 'false'.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -223,6 +223,10 @@ class Loader
             $parts = explode(' #', $value, 2);
             $value = trim($parts[0]);
 
+			if ('false' == $value) { # getenv() returns a string, 'false' is not falsey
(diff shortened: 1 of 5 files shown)


GrahamCampbell:
Sorry but this is a breaking change, and can't be accepted. It also prevents someone from using the string `false`.

PR #179, opened 2016-05-06
Support disabling of trimming of values wrapped in double quotes

### LINKED ISSUE #162: Trimming of string literals
Due to double trimming of values at Loader::sanitiseVariableValue, string literals are being trimmed; that is, `HERP = "     please don't"` is forced to be `please don't`.

It didn't cause problems as of yet, because everyone is accustomed to space and tabs being unsafe characters, but no one should have to rely on this behaviour.

GrahamCampbell:
What do you make of this?

Note that I've set the default option to not change the current behaviour for BC reasons.

---
Closes #162.

### DIFF
--- a/tests/Dotenv/DotenvTest.php
+++ b/tests/Dotenv/DotenvTest.php
@@ -47,6 +47,14 @@ class DotenvTest extends PHPUnit_Framework_TestCase
         $this->assertEmpty(getenv('CNULL'));
     }
 
+    public function testDotenvLoadsEnvironmentVarsWithoutTrimmingQuoted()
+    {
+        $dotenv = new Dotenv($this->fixturesFolder, 'commented.env', false);
+        $dotenv->load();
+        $this->assertSame('bar', getenv('CFOO'));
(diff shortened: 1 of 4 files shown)


tests/Dotenv/DotenvTest.php
```diff
@@ -47,6 +47,14 @@ public function testCommentedDotenvLoadsEnvironmentVars()
         $this->assertEmpty(getenv('CNULL'));
     }
 
+    public function testDotenvLoadsEnvironmentVarsWithoutTrimmingQuoted()
+    {
+        $dotenv = new Dotenv($this->fixturesFolder, 'commented.env', false);
+        $dotenv->load();
+        $this->assertSame('bar', getenv('CFOO'));
```
GrahamCampbell: note how this variable should not have the leading spaces because it wasn't quoted, but the other one will have trailing spaces because it was quoted

GrahamCampbell:
Tests are failing. @pawelabrams Looks like PHP is trimming them itself anyway?

PR #181, opened 2016-05-18
Load the sub "env" file in the "env" file.

weiyongsheng:


GrahamCampbell:
Thanks. I'm not sure I really like the idea though. What's the use case?

PR #183, opened 2016-06-14
Avoid code duplication

xxnoobmanxx:


src/Dotenv.php
```diff
@@ -80,6 +76,16 @@ protected function getFilePath($path, $file)
 
         return $filePath;
     }
+    
+    /**
+     * Load data from '.env'
```
GrahamCampbell: This description is not correct, but I can fix it.

src/Dotenv.php
```diff
@@ -80,6 +76,16 @@ protected function getFilePath($path, $file)
...
+     * @return array
+     */
+    private function getLoaderData($overload = false)
+    {
+        $this->loader = new Loader($this->filePath, !$overload);
+        return $this->loader->load();
```
GrahamCampbell: I'll need to restore the original code style here.

src/Dotenv.php
```diff
@@ -80,6 +76,16 @@ protected function getFilePath($path, $file)
 
         return $filePath;
     }
+    
+    /**
+     * Load data from '.env'
+     * @return array
+     */
+    private function getLoaderData($overload = false)
```
GrahamCampbell: this will probably need to be protected rather than private

GrahamCampbell:
Thanks, I'll apply the remaining fixes.

PR #190, opened 2016-08-04
helper function to create a new environment

jordanbardsley7:


vlucas:
Sorry, but I don't want to pollute the global namespace by defining a function. A better option would be to create a static method on the Dotenv class that could be used instead.

jordanbardsley7:
That sounds like a better idea, ill create a PR later!

PR #191, opened 2016-08-05
statically create a new environment

jordanbardsley7:
Changes made from PR https://github.com/vlucas/phpdotenv/pull/190.


GrahamCampbell:
Just new up the class? We don't need to add extra functions?

Issue #104, opened 2015-07-15
Detect and convert types


oscarotero:
In my `.env` file I have values like:

```
APP_DEBUG=true
APP_DEV=false
```

But the type of these values is always string:

```php
var_dump(getenv('APP_DEBUG')); // string(4) "true"
var_dump(getenv('APP_DEV')); // string(5) "false"
```

I think the values like `true`, `false`, `null` (or even numbers), without quotes, should be converted to the correct type (boolean, integers, etc), because in cases like this, `APP_DEV` returns `"false"` so it's evaluated as `true`:

```php
if (getenv('APP_DEV')) {
    // do something
}
```

GrahamCampbell:
We do this don't we?

GrahamCampbell:
We have tests that prove this works.

oscarotero:
Currently it returns `false` if the variable is not defined, but I don't get the values `true` or `false` converted to booleans.
In the tests I can see [this fixture](https://github.com/vlucas/phpdotenv/blob/master/tests/fixtures/env/mutable.env) and [the expected value is string](https://github.com/vlucas/phpdotenv/blob/master/tests/Dotenv/DotenvTest.php#L219), not boolean.

GrahamCampbell:
What version?

GrahamCampbell:
Hmmm.

oscarotero:
The latest dev version (2.1@dev)

GrahamCampbell:
NB, 2.0.x is the latest version. I don't recommend using an unreleased version as it's still subject to change.

oscarotero:
Yes, you're right :smile: 2.0.1 does the same.

vlucas:
@GrahamCampbell These are converted in Laravel, but not in phpdotenv itself. The only reason this is not done in phpdotenv is because I originally tried to keep it as close to shell as possible, and in shell ENV variables, everything is a string. I am not opposed to adding this functionality in, though - and we can probably even just grab the code from Laravel that does this and put it in phpdotenv.

oscarotero:
Maybe, a way to keep backward compatibility is make this behaviour optional. For example:
```php
$dotenv->load(); // do not convert types
$dotenv->load(true); // convert types
```

oscarotero:
Here's how laravel handle this: https://github.com/laravel/framework/blob/8c614010648bffb8bc24c3b532db08619e9a5983/src/Illuminate/Foundation/helpers.php#L626

[...]

PR #192, opened 2016-08-28
Boolean validation

snellingio:
Wanted the option to assert boolean values. 

I think isBoolean makes more sense than allowedValues for true and false.

This is useful when setting a debug env flag, etc.


snellingio:
Looks like this may be out of scope due to #104

GrahamCampbell:
Values will never be boolean. They're always strings?

PR #197, opened 2016-10-01
Added the Valiator method isBoolean

sephedo:


GrahamCampbell:
Thanks for the PR. Our variables are always strings though sent they?

sephedo:
Yes all values are in the format of strings this method ensures they are in the format accepted as booleans - Yes, No, 1, 0, On and Off

danhunsaker:
Has some weird indentation, but otherwise it looks sound.

jpuck:
whoops, looks like the tests are not compatible with ye olde php 5.3
I'll update that to use the formal `array()` instead of `[]`

jamesj2:
It's been a year. Any possibility of this getting merged?

vlucas:
This looks pretty solid to me. Has tests and will not affect anyone if they are not using the boolean validator.

PR #198, opened 2016-10-04
Make it easier to swap out the loader, to allow for a custom loader

nathanmac:
Small PR which just moves the loader to a place where it can be easily overwritten to allow for custom loaders and adds a `LoadedInterface`

My particular use-case for this concerns some additional pre-processing steps necessary to allow the `.env` file to be readable, therefore requiring a custom loader with a minor addition to the existing loader, this would make it easier/cleaner to overwrite the loader in the `Dotenv` class.

### DIFF
--- a/src/Dotenv.php
+++ b/src/Dotenv.php
@@ -20,7 +20,7 @@ class Dotenv
     /**
      * The loader instance.
      *
-     * @var \Dotenv\Loader|null
+     * @var \Dotenv\LoaderInterface|null
      */
     protected $loader;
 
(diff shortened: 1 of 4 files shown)


danhunsaker:
Doesn't look like this change actually lets you supply your own loader instance at any point...  It just adjusts the abstraction a bit to call a method rather than directly construct anything inline.

nathanmac:
That's correct the idea was to remove the double instantiation so the get loader method could be overwritten in a single place, this appears to have been addressed in another PR vlucas/phpdotenv#215.

GrahamCampbell:
@nathanmac NB Futher improvements have been made in 3.x series. :)

Issue #196, opened 2016-09-09
A way to store values in $_ENV but not $_SERVER


benjaminkohl:
It was brought to our attention that the $_SERVER super global's contents are dumped to the browser console when a particular CMS is in development mode. While we only do this on the dev site, we'd still like to keep our DB credentials out of the browser console.

Is there a way to make dotenv put/get from $_ENV entirely and not make any changes to $_SERVER?

stedaniels:
See https://github.com/vlucas/phpdotenv/pull/200 this would mitigate this potential vulnerability in the CMS.  Hopefully it'll get merged.

GrahamCampbell:
Done in https://github.com/vlucas/phpdotenv/pull/300. See https://github.com/vlucas/phpdotenv#loader-customization for an example.

Issue #163, opened 2016-02-18
Allow read-only mode


barryvdh:
There has been some discussion about using putenv() and setting the $_ENV/$_SERVER vars, related to security and threads, eg #76 and #124, https://github.com/laravel/framework/pull/8187 and Twitter: https://twitter.com/weltling/status/700036432761659392 etc.

Those commonly revolve about setting (sensitive) data in the Environment variables. While usage in production is discouraged (through a small note in the end of the readme), it doesn't mention anything about security issues (like leaking data to other processes or turning up in dumps)

My guess is that this library is mainly used to READ files from both environment vars and the .env file, but for many cases doesn't really care about putting stuff back. So why not add a 'read only' mode:
- Loading vars loads all data in an internal cache, instead of putting them in Env vars
- Finding a var will use the cache as first source (eg `dotenv('DB_PASS')` instead of `getenv('DB_PASS')`)

(Similar to what @progmars suggested in https://github.com/vlucas/phpdotenv/issues/76#issuecomment-119975430 but also not using the `putenv` anymore).

This could be an optional flag, just like immutable.

barryvdh:
See this issue here about `putenv`: https://bugs.php.net/bug.php?id=71607 and the commenter here: http://php.net/manual/en/function.putenv.php#8496

So the issue has 2 parts:
 - `putenv` leaks vars to system processess, which could be unwanted.
 - $_ENV/$_SERVER can be printed by accident on wrongly configured systems.

Removing `putenv()` would prevent the first issue, but if you want to avoid the second case also, we need to remove that as well.

tuupola:
> putenv leaks vars to system processess, which could be unwanted.

This is how environment works on *NIX systems. It is a feature, not a bug.

There is nothing especially dangerous in storing sensitive information into environment than compared storing it somewhere else. Yes, if you dump your environment the information can be seen. In the same way if you dump stack trace, for example `PDOException`, the database credentials will be visible. It does not matter where you originally store them.

> While usage in production is discouraged

This is about using an `.env` file in production because it causes extra disk access, and not about using environment variables in production. Latter is quite common and is actually [encouraged](http://12factor.net/config).

weltling:
@tuupola your conclusions are wrong in a lot of ways.

- stable system is not about a potentially compromised system
- dumps you mention are not relevant in production
- "lib to use in dev only" is not about a "factor III in the 12 factor concept" or a general purpose lib, or the doc is wrong

Furthermore
[...]

PR #200, opened 2016-10-10
Allow configuration flags to be set within individual .env files

adam-prickett:


stedaniels:
@vlucas is anything needed to get this merged?  Or do you just not want to do this kind of thing and you're being polite by not saying? :-)  Being too busy too look at it is totally a valid reason.  (I've got three kids under three!)

cronfy:
I do not like the idea that .env file dictates to webserver where to store environment variables. It is also not flexible to put these settings in the .env file.

Why not just have a `load()` equivalent, that only returns array of env variables without touching `$_ENV/$_SERVER/putenv()`, e. g. `read()`? (also see #163).

tuupola:
I am against merging this patch. One should be able to do `$ source .env` and expect the shell to have same environment variables as PHP has after calling `$dotenv->load();`.

vlucas:
I don't like controlling phpdotenv itself with environment variables that it loads. This seems like something that should be in the PHP code or config instead.

schmunk42:
@vlucas I just came across the same issue, like https://github.com/vlucas/phpdotenv/issues/196; several frameworks also log $_SERVER which is populated with ENV variables.

Sure, $_SERVER contains critical values by default, but it should be up to the developer if he wants to *populate* $_SERVER or not.

The current hacky way is to sanitize $_SERVER manually.

cronfy:
I like the way it is solved here: https://github.com/josegonzalez/php-dotenv

```
$Loader = (new josegonzalez\Dotenv\Loader('path/to/.env'))
              ->parse();
```

Then array of options can either be returned via `toArray()` or populate env via `toEnv()`, but it is the developer who decides, where env variables go.

GrahamCampbell:
Replaced by https://github.com/vlucas/phpdotenv/pull/300. See https://github.com/vlucas/phpdotenv#loader-customization for an example.

PR #201, opened 2016-10-12
added support for null and boolean values

opichon:
To specify a null value, use the keyword `null`. To specify the boolean true value, use the keywords `true`, `yes,` `on`. To specify the boolean false value, use the keywords `false`, `no`, `off`. 

The keywords must be used without quotes.

If used with quotes, the actual string will be used instead.

### DIFF
--- a/src/Dotenv.php
+++ b/src/Dotenv.php
@@ -29,8 +29,6 @@ class Dotenv
      *
      * @param string $path
      * @param string $file
-     *
-     * @return void
      */
     public function __construct($path, $file = '.env')
     {
--- a/src/Exception/ExceptionInterface.php
+++ b/src/Exception/ExceptionInterface.php
@@ -7,5 +7,4 @@ namespace Dotenv\Exception;
  */
 interface ExceptionInterface
 {
-    //
 }
--- a/src/Exception/InvalidCallbackException.php
+++ b/src/Exception/InvalidCallbackException.php
@@ -9,5 +9,4 @@ use InvalidArgumentException;
  */
 class InvalidCallbackException extends InvalidArgumentException implements ExceptionInterface
 {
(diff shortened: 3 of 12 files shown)


GrahamCampbell:
Thanks, but this has already been discussed, and strings are fine. You just need to interpret them how you want when you use them.

opichon:
Sorry, I could not see the discussion in the issues. Where can I find it? I'd be interested in understanding the reasons why this aspect was decided against. I do notice that getenv always returns a string or a literal value, (eg `''` for null and `0` or `1` for booleans.

In any case, may I insist? To "interpret" the values is not very convenient, unfortunately. This is our use case:

We use symfony 2.8|3.1.

app config is stored in `app/config/config.yml` with entries such as:
``` yaml
# Swiftmailer Configuration
swiftmailer:
    transport:  "%mailer_transport%"
    host:       "%mailer_host%"
    username:   "%mailer_user%"
    password:   "%mailer_password%"
    encryption: %mailer_encryption%
    port:       "%mailer_port%"
    auth_mode: %mailer_auth_mode%
```

The parameters (eg `%mailer_encryption%`) are defined in our `.env` file as :

```
SYMFONY__MAILER_ENCRYPTION=null
```

Symfony automatically transforms evn vars prefixed with `SYMFONY__`` into the relevant parameter.

Unfortunately this does not allow for strings to be interpreted as null or boolean values. Note that using quotes or not makes no difference.

Now we could try and fix the symfony side of things, but it seems a lot easier to fix dotenv.

opichon:
I could only find this MR: https://github.com/vlucas/phpdotenv/pull/177 which is not quite the same as mine.

Could you at least point me to the relevant discussion so that I can understand the issue?

PR #209, opened 2016-11-18
Loader: streamline isComment()

jrchamp:
Skip using strpos() because we already know the string position that needs to be checked.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -147,7 +147,8 @@ class Loader
      */
     protected function isComment($line)
     {
-        return strpos(ltrim($line), '#') === 0;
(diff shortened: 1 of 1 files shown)


jrchamp:
Would have done it in one line, but Travis complained about PHP 5.3.

GrahamCampbell:
This fails for the empty string.

GrahamCampbell:
The current code works correctly, however this code will throw an error whenever only whitespace, or the empty string is passed.

jrchamp:
Technically, an empty string would never be passed to this function the way it is set up.

Please reopen this pull request so I can at least fix the tests.

Thank you.

PR #210, opened 2016-11-18
Dotenv: Unused Loader instance in __construct

jrchamp:
loadData() will always create a Loader instance, so creating one in Dotenv::__construct() is only useful for Validate.

Hopefully travis agrees.

### DIFF
--- a/src/Dotenv.php
+++ b/src/Dotenv.php
@@ -35,7 +35,9 @@ class Dotenv
     public function __construct($path, $file = '.env')
     {
         $this->filePath = $this->getFilePath($path, $file);
-        $this->loader = new Loader($this->filePath, true);
+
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Hmm. Not sure this is the best fix actually.

GrahamCampbell:
I think we have a more fundamental design problem here.

jrchamp:
Completely agree that this is not the best fix. This only makes it more clear what is happening whereas right now, it's somewhat misleading. Feel free to close this out when you have a better solution. If you can, please link to that better solution here.

vlucas:
Just merged #215 which fixes this problem in `loadData`.

jrchamp:
Thanks @vlucas, closing.

PR #211, opened 2016-11-26
no return comment __construct

nicolas-wallerand:


GrahamCampbell:
We prefer to always put the return annotation in order to differentiate between actually returning nothing, and just forgetting to document the return type.

nicolas-wallerand:
return type does not match

A number of parameters, their names or types  (if any) in PHPDoc comment do not match the respective function or method declaration.

but https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#715-return

[functions and methods without a return value: the @return tag MAY be omitted here, in which case an interpreter MUST interpret this as if @return void is provided.]

GrahamCampbell:
The return type does match. That function has do return statement, so it doesn't return anything.

I'm not sure what you mean by quoting those docs? They don't back up your point?

GrahamCampbell:
Locking since this discussion isn't useful.

PR #212, opened 2016-11-28
Loader: streamline isComment() with whitespace test

jrchamp:
Note: `isset()` is a language construct and therefore does not incur the overhead of a true function call.

Supersedes #209.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -147,7 +147,9 @@ class Loader
      */
     protected function isComment($line)
     {
-        return strpos(ltrim($line), '#') === 0;
+        $line = ltrim($line);
+
+        return isset($line[0]) && $line[0] === '#';
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Loader.php
```diff
@@ -147,7 +147,8 @@ protected function readLinesFromFile($filePath)
      */
     protected function isComment($line)
     {
-        return strpos(ltrim($line), '#') === 0;
+        $line = ltrim($line);
```
GrahamCampbell: Please add a newline after this.

src/Loader.php
```diff
@@ -147,7 +147,8 @@ protected function readLinesFromFile($filePath)
...
+        return (isset($line[0]) && $line[0] === '#');
```
GrahamCampbell: No need for the outer set of brackets please.

jrchamp:
Changes applied as requested.

GrahamCampbell:
Thanks.

PR #215, opened 2016-12-14
remove double instantiation of loader class

arifmahmudrana:
add set & get immutable method in loader class
add tests for set & get immutable method


vlucas:
Thanks for this. 👍

GrahamCampbell:
I'm not sure we should allow modification of the immutability of the loader. :-1: on this. Kinda defeats the point of making it immutable if we allow that flag to be mutated.

PR #220, opened 2017-01-31
Loader: avoid wasteful strpbrk, array_key_exists

jrchamp:
`array_key_exists()` could return `true` when the value is set to `null`, the same result is considered to be a failure in the `default:` case. This may prevent a value in `$_SERVER` from being used.

`strpbrk()` is a rarely used function that behaves like `strstr()` and returns a substring on success. Given that we are only checking a single character for specific values, `isset()` (a language construct that does not incur standard function overhead) is more suitable for performance and legibility.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -312,7 +312,7 @@ class Loader
      */
     protected function beginsWithAQuote($value)
     {
-        return strpbrk($value[0], '"\'') !== false;
+        return isset($value[0]) && ($value[0] === '"' || $value[0] === '\'');
     }
 
     /**
@@ -325,9 +325,9 @@ class Loader
     public function getEnvironmentVariable($name)
     {
         switch (true) {
-            case array_key_exists($name, $_ENV):
+            case isset($_ENV[$name]):
                 return $_ENV[$name];
-            case array_key_exists($name, $_SERVER):
+            case isset($_SERVER[$name]):
                 return $_SERVER[$name];
             default:
                 $value = getenv($name);


src/Loader.php
```diff
@@ -325,9 +325,9 @@ protected function beginsWithAQuote($value)
     public function getEnvironmentVariable($name)
     {
         switch (true) {
-            case array_key_exists($name, $_ENV):
+            case isset($_ENV[$name]):
```
GrahamCampbell: this is actually not correct

jrchamp: What's not correct? It works correctly and fixes two test cases for me over the current code.

I ran my test case through 3v4l to make sure it worked on all versions: https://3v4l.org/ZP3gR

> Error for orig method - expected: 'both_value_server_not_null'; found: NULL
> Error for orig method - expected: 'all_value_both_null_getenv'; found: NULL

rquadling: Wouldn't falsy values mean the default value always gets used?

jrchamp: @rquadling The only value that `isset()` does not like is `null`. Other falsy values, such as `false`, `0`, `''` (the empty string) and `'0'` (the string 0) are considered to be "set". On the other hand, `empty()` and implicit/automatic boolean conversion do not like those values: https://secure.php.net/manual/en/language.types.boolean.php

GrahamCampbell: The change in behaviour is now if the env variable was null, then the server variable gets incorrectly returned instead.

GrahamCampbell:
Please send just the beginsWithAQuote change.

src/Loader.php
jrchamp: The current behavior is "when $_ENV[$name] is null, return null" which is interpreted as "not found" in functions like `resolveNestedVariables()`. It seemed like the desired behavior should be to try each data source in the order provided to return a "found" value rather than exit earlier with a "not found" value. From what you said, it seems that `null` is a valid "found" value, which I find confusing, but can accept.

GrahamCampbell: Still a breaking change though as far as I can see.

jrchamp:
Closing this issue. @GrahamCampbell has confirmed that null values within `$_ENV` and `$_SERVER` are valid return values from `getEnvironmentVariable()`

PR #222, opened 2017-02-03
Added clarity to usage with Composer

tommarmstrong:


GrahamCampbell:
👎 Not normally correct. Just confusing.

fabswt:
Shouldn't the README state:

    php composer.phar require--dev vlucas/phpdotenv

Instead of 

    php composer.phar require vlucas/phpdotenv

?

GrahamCampbell:
Not dev.

fabswt:
And why not then?

The README says:

> phpdotenv is made for development environments, and generally should not be used in production

I'm using Heroku's config vars for production, why would I need to install phpdotenv as `require` instead of `require-dev`?

GrahamCampbell:
> phpdotenv is made for development environments, and generally should not be used in production

That readme line is outdated, and needs fixing. You should not be executing phpdotenv, or indeed, any other similar library using php-fpm, or related. Instead, you should do something like what Laravel's config system does, and cache environment variables into computed config files, using a console command.

PR #223, opened 2017-02-03
Loader: use isset instead of strpbrk

jrchamp:
Supercedes #220:

`strpbrk()` is a rarely used function that behaves like `strstr()` and returns a substring on success. Given that we are only checking a single character for specific values, `isset()` (a language construct that does not incur standard function overhead) is more suitable for performance and legibility.


GrahamCampbell:
Thanks. 👍

PR #224, opened 2017-02-09
Adding option to ignore .env file requirement

dpires:
The Ruby dotenv implementation has 2 load methods (load, load!). The load method ignores a missing .env file for the case where you are in a non-development environment and have your environment variables set by some other method.

I would like the same options for phpdotenv.

For backwards compatibility I have added an optional flag to load() to ignore the requirement of having a .env file. By default it is true.

### DIFF
--- a/src/Dotenv.php
+++ b/src/Dotenv.php
@@ -41,11 +41,17 @@ class Dotenv
     /**
      * Load environment file in given directory.
      *
+     * @param bool $requireDotEnv
+     *
      * @return array
      */
-    public function load()
+    public function load($requireDotenv = true)
     {
-        return $this->loadData();
+        if ($requireDotenv) {
+            return $this->loadData();
+        }
+
+        return array();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Dotenv.php
```diff
@@ -41,11 +41,16 @@ public function __construct($path, $file = '.env')
...
      *
+     * @param bool $requireDotEnv
+     *
      * @return array
      */
-    public function load()
+    public function load($requireDotenv = true)
     {
-        return $this->loadData();
+        if ($requireDotenv)
```
jpuck: missing braces per [PSR-2](http://www.php-fig.org/psr/psr-2/#if-elseif-else)

svjacob:
yes please. 💯

GrahamCampbell:
Sorry, didn't see this. #242 has been merged now, since it fits the BC promise of this library better. Adding a parameter to a public function in a non-final class is technically a BC break in PHP.

PR #226, opened 2017-02-27
Fix incorrect nesting documentation

calebporzio:
`{$variable}` does not work, but `${variable}` does. The docblock for the respolveNestedVariables method is incorrect.


GrahamCampbell:
🍻

PR #227, opened 2017-03-05
Docker secrets

opichon:
Added support for docker secrets. See relevant paragraph added to the README file for the rationale for this PR.


vlucas:
I think the convention of automatically loading and reading a file just by the ENV name `*._FILE` would be very problematic given how widely used phpdotenv is. There are very likely to be people already using phpdotenv that have ENV names that would match that, that could then cause very big and unexpected surprises when they upgrade if this is merged.

I don't see a way phpdotenv could ever accept a special rule like this around naming conventions. it would have to be done with a separate function call or similar, if it even belongs in this project at all.

PR #240, opened 2017-07-17
Create helpers.php

wirwolf:


GrahamCampbell:
Thanks for the PR, but I'd rather not lift code from Laravel and put it in here. :)

wirwolf:
I use another framework i must added this file in each new project

vlucas:
@wirwolf I would recommend creating and installing a separate composer package for this. We would like to keep DotEnv's core responsibilities to only .env file loading/parsing.

PR #241, opened 2017-07-29
Update README.md

morrislaptop:
Link to repo to validate required vars from .env.example

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,9 @@ If any ENV vars are missing, Dotenv will throw a `RuntimeException` like this:
 One or more environment variables failed assertions: DATABASE_DSN is missing
 ```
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Thanks, but I'd rather not recommend extra packages. If there's a useful feature people want, I'd rather have it contributed here. ;)

morrislaptop:
I didn't think it belonged here with the existing required validation. 

But no worries, thanks for getting back to me so fast.

GrahamCampbell:
Depends if there's a demand for it. If there are lots of other community packages, I'd be up for adding a wiki page for them, like pysh does: https://github.com/bobthecow/psysh/wiki/Integrations.

PR #242, opened 2017-07-31
Ignore missing file

endorama:
Hello everyone, first of all thanks for this project.

I'd like to introduce ( as in the Ruby counterpart ) the skip-on-missing-file feature, particularly handy for dev/test environments.

This implementation introduce this feature for `load` and `overload` and provides a new `force_load` method to explicitly retain the current behaviour ( throw on missing ).

### DIFF
--- a/src/Dotenv.php
+++ b/src/Dotenv.php
@@ -2,6 +2,8 @@
 
 namespace Dotenv;
 
+use Dotenv\Exception\InvalidPathException;
+
 /**
  * This is the dotenv class.
  *
@@ -46,6 +48,21 @@ class Dotenv
     public function load()
     {
         return $this->loadData();
+    }
+
+    /**
+     * Load environment file in given directory, suppress InvalidPathException.
+     *
+     * @return array
+     */
+    public function safeLoad()
+    {
+        try {
+            return $this->loadData();
+        } catch (InvalidPathException $e) {
+            // suppressing exception
+            return array();
+        }
     }
 
     /**
--- a/tests/Dotenv/DotenvTest.php
+++ b/tests/Dotenv/DotenvTest.php
@@ -24,6 +24,12 @@ class DotenvTest extends PHPUnit_Framework_TestCase
         $dotenv->load();
     }
 
+    public function testDotenvSkipsLoadingIfFileIsMissing()
+    {
+        $dotenv = new Dotenv(__DIR__);
+        $this->assertEmpty($dotenv->safeLoad());
+    }
+
     public function testDotenvLoadsEnvironmentVars()
     {
         $dotenv = new Dotenv($this->fixturesFolder);


src/Dotenv.php
```diff
@@ -40,22 +40,41 @@ public function __construct($path, $file = '.env')
 
     /**
      * Load environment file in given directory.
+     * If .env file is missing, skip loading and return false.
      *
-     * @return array
+     * @return boolean|array
```
GrahamCampbell: This is a major breaking change I'm afraid.

src/Dotenv.php
```diff
@@ -40,22 +40,41 @@ public function __construct($path, $file = '.env')
...
+        return false;
     }
 
     /**
      * Load environment file in given directory.
+     * If .env file is missing throw an exception.
      *
      * @return array
      */
+    public function force_load()
```
GrahamCampbell: Please use camelCase. :)

src/Dotenv.php
```diff
@@ -40,22 +40,41 @@ public function __construct($path, $file = '.env')
...
      * Load environment file in given directory.
+     * If .env file is missing, skip loading and return false.
      *
-     * @return array
+     * @return boolean|array
      */
     public function load()
     {
-        return $this->loadData();
+        if (file_exists($this->filePath)) {
```
GrahamCampbell: There is a race condition here. Better to catch the exception.

endorama: I'm sorry, but I fail to see the race condition.

From [php doc](http://php.net/manual/en/function.file-exists.php):
```
Returns TRUE if the file or directory specified by filename exists; FALSE otherwise.
```

What is the possible exception? Do you think would be better to use [`is_readable`](http://php.net/manual/en/function.is-readable.php)?

src/Dotenv.php
endorama: Yes sorry! I'm going to return an empty array.

endorama:
@GrahamCampbell thank you for reviewing the code.

I made the required changes!

src/Dotenv.php
```diff
@@ -40,22 +40,41 @@ public function __construct($path, $file = '.env')
 
     /**
      * Load environment file in given directory.
+     * If .env file is missing, skip loading and return an empty array.
      *
      * @return array
      */
     public function load()
     {
+        if (file_exists($this->filePath)) {
```
GrahamCampbell: There is still a race condition here. The file could be deleted, or be inaccessible between this test, and actually getting the file. Please leave the existing behaviour of load as is, and have `safeLoad` be the method that suppressed exceptions.

endorama:
@GrahamCampbell I refactored the code as you requested.

But reading the load function in `Loader` ( [lines 75 and 78](https://github.com/endorama/phpdotenv/blob/5d0709193c68dcbc70479e88b74cd044e37191ff/src/Loader.php#L75-L78) ) my understanding is that the same race condition exists. Am I wrong?

Thank you.

svjacob:
dupe of https://github.com/vlucas/phpdotenv/pull/224? 👀

endorama:
@svjacob the main idea behind this PR was to be more similar to the ruby `dotenv` library.

The final implementation after some reviews by @GrahamCampbell is really similar to the one in the #224. ( Code has been rebased, so unfortunately is not possible to look at that implementation anymore )

Personally I would avoid a flag that changes method behaviour :) but I'm open to alternatives.

GrahamCampbell:
> @svjacob the main idea behind this PR was to be more similar to the ruby dotenv library.

I'd rather maintain BC than mimic the ruby API.

GrahamCampbell:
This PR looks very good now. 👍 

One other thing. Is there any chance you could add `@throws` annotations to the functions that may raise the exception. This is really a documentation bug on our part. :)

GrahamCampbell:
Will merge this as is, and make an issue for the other thing.

Thank you! 🍻

endorama:
Thank you!

I'll follow the related issue and make a PR for it tomorrow ( I'm in CEST
timezone ).

endorama:
@GrahamCampbell Today I noticed that the 2.4.0 release does not contain this code. Is this expected?

PR #245, opened 2017-09-13
Allow for includes

etelford:
I originally created this as an extension to the doting package, but figured it would be worth a shot putting this in as a PR.

This allows for using include files in you `.env`. I find this helpful in large projects when your `.env` can get quite large and out of hand.

Example:

```shell
# your root .env
S3_INCLUDE=s3.env

# s3.env
BUCKET=mybucket
KEY=sup3rs3cr3tk3y!
```

And to use these, you would do:

```php
getenv('S3_BUCKET');
```

All other features that work with `.env` variables, like comments, nesting, etc. work inside include files.

You can also have include files inside your include files if you're daring enough.


etelford:
It looks like Travis had an issue building. I'm going to close and re-PR to force another build.

PR #246, opened 2017-09-13
Allow for includes

etelford:
_I'm resubmitting this PR so Travis can re-run tests._

I originally created this as an extension to the doting package, but figured it would be worth a shot putting this in as a PR.

This allows for using include files in you `.env`. I find this helpful in large projects when your `.env` can get quite large and out of hand.

Example:

```shell
# your root .env
S3_INCLUDE=s3.env

# s3.env
BUCKET=mybucket
KEY=sup3rs3cr3tk3y!
```

And to use these, you would do:

```php
getenv('S3_BUCKET');
```

All other features that work with `.env` variables, like comments, nesting, etc. work inside include files.

You can also have include files inside your include files if you're daring enough.


carstenwindler:
I think this is a very useful feature!

Obviously PHP 5.3 is no longer supported on Ubuntu Trusty (https://github.com/travis-ci/travis-ci/issues/8219). Maybe it's time to drop PHP 5.3 support anyway? @vlucas

etelford:
It looks like several Pos have been open for quite a long time. Is @vlucas still actively maintaining?

GrahamCampbell:
Thanks for the PR. I'm in two minds as to if this feature should be included. I'll sleep on it. :)

vlucas:
I am mixed on this as well. I think for this to work, it needs to have some sort of constant variable name that would not possibly be used normally, like `__DOTENV_INCLUDE=s3.env`. Re-using the same variable name would also avoid ENV variable pollution when including multiple files, because still only a single extra variable would be set.

purplekrayons:
While I understand the OP's perspective, I personally disagree with the idea of supporting includes in environment files as you've described with this PR. I also disagree with this related pull request https://github.com/vlucas/phpdotenv/pull/288 for the same reason; they both promote the use of multiple environment files.

I prefer the concept this idea package promotes, of having a single environment variable, and a single env sample file to indicate what people should expect.

Allowing someone to implement numerous environment files that I would then need to individually compare sample files, feels like it could become tedious and introduces unnecessary complexities, particularly within large teams. The fact that there is only one env file, makes it inherently easy to on-board new team members to projects, as opposed to saying:

> Here you go new person, you'll need these 12 environment files I've setup.

I'm also not sure I agree with the mentality that you even need multiple environment variables. Typically, these are values which should rarely change, at least in a production environment anyway. It's one of those things where you can set it and forget it. We must also consider the added overhead, albeit it minor, from the functions required to implement this. I believe this package should remain as lightweight as possible.

GrahamCampbell:
Yeh, I'm inclined to say this is likely too complex. Env files should not be confused for configuration files. They should be mainly used for setting secrets, and not everything under the sun. :)

GrahamCampbell:
Thank you for the PR though. :)

PR #249, opened 2017-11-15
Fix PHP 5.3 on Travis CI

gabriel-caruso:
I've changed `PHP 5.3` on `Travis CI` to run on `Precise` instead of `Trusty`. This fixes our tests.


vlucas:
Thank you! I was just going to drop it :)

PR #256, opened 2017-12-31
Update license year

m1guelpf:
TO DO:

- [x] Wait to the next year (100% done, depends on timezone)
- [ ] Merge!

:octocat:


GrahamCampbell:
Thanks, but the copyright year needs to remain at the earliest year to be valid.

GrahamCampbell:
(or will need to be a year range, which needs maintaining every year, so easiest to not do a range)

PR #260, opened 2018-02-01
Docker Secrets Support

publiux:
This pull request implements support for Docker secrets. As containerization becomes more prevalent, it is often necessary to avoid committing some values in .env files to source control, or even using them to create containers.

This proposal creates a new format for .env files where an environment name can end with two pipes (||) and have no value in the file. In this instance, phpdotenv will read the Docker secret file with the same name (without the pipes) and load the value into the environment variables.


publiux:
Seeing as though some pull requests have lingered here for many months without resolution, I've gone ahead and create a fork of this project at [publiux/phpdotenv](https://github.com/publiux/phpdotenv) with this change. If this PR is merged, that project will be deprecated. I have also put the project on packagist for use by anyone that needs Docker secret support through phpdotenv.

src/Loader.php
```diff
@@ -187,6 +215,18 @@ protected function looksLikeSetter($line)
...
+	/**
+     * Determine if the given line looks like it's setting a variable from a Docker secret.
+     *
+     * @param string $line
+     *
+     * @return bool
+     */
+    protected function looksLikeDockerSecret($line)
+    {
+    	return strpos($line, '||') !== false;
```
AubreyHewes: won't this fail with a line that contains a variable with the delimiter `||` i.e. `test="testing||this"` ?
possibly change the test for a line ending in `$docker` ? this removes any false positives; any envvar ending in `$docker` is read from the the other file...

src/Loader.php
```diff
@@ -374,6 +414,48 @@ public function setEnvironmentVariable($name, $value = null)
...
+     * - $_SERVER.
+     *
+     * The environment variable value is stripped of single and double quotes.
+     *
+     * @param string      $name
+     * @param string|null $value
+     *
+     * @return void
+     */
+    public function setDockerEnvironmentVariable($name)
```
AubreyHewes: keep it dry; this should be incorporated withing `setEnvironmentVariable` . i.e. therein a test `isDockerEtc` (sic)

src/Loader.php
publiux: The delimiter must be at the end, so having "testing||this" wouldn't invoke the docker secret.

publiux:
Understandable you might not understand it. Docker secrets reside in the /run/secrets directory by default. The name of the secret appears as a file in that directory. This PR allows you to have a .env file that contains variables and values you can commit to a repository or can deploy on a server as well as instructing this package to retrieve certain secrets that are not safe to commit to a .env file from the docker secrets feature.

For example, AWS keys can be docker secrets, but telling an application the IP of the SQL server isn't much of a secret. This PR allows you to mix both into one .env file.

src/Loader.php
AubreyHewes: it would as it is a a check that the `line` _contains_ a `||`.  The line can be something like `test=this` . `this` can contain a `||`

i.e. `test="this||"`

AubreyHewes:
Ah ok, I understand what you mean with the files within `/run/secrets`, i.e. `/run/secrets/X` apologies on that.

Though this solution does not belong in this project. It is too specific.

A better PR is to expand the current interface to allow adding vars before loading `.env` files. This ensures your vars are leading.

i.e.

```php
  $dotenv = new Dotenv\Dotenv();
  $dotenv->add(getMyVars()); // supersede matching `.env` vars
  $dotenv->load();
```

And/Or PR with a plugin interface and make a plugin

```php
  $dotenv = new Dotenv\Dotenv();
  $dotenv->plugin(new \MyVars()); // supersede matching `.env` vars
  $dotenv->load();
```

btw; the line tests are broken and still need to be fixed.

src/Loader.php
GrahamCampbell: Can someone please implement this change, so we can move forward with this PR?

Issue #234, opened 2017-05-31
trim the last spaces in the values of prisoners in quotes


ikee87:
Hi!
in file .env i have line
```
XI_SERVICE_PASSWORD="MY PASSWORD WHIT LAST SPACE "
```
In result i have string whit out last space... =( fix this.

https://github.com/vlucas/phpdotenv/blob/master/src/Loader.php#L257

rgehan:
Is this really bad behavior ?
Test cases in [tests/Dotenv/DotenvTest.php]() seem to imply it is like this on purpose

Korbeil:
Since it's explicit value between double quotes, it should not be trimed.

GrahamCampbell:
Tripping will be stopped in v3.0, implemented in https://github.com/vlucas/phpdotenv/pull/273. :)

PR #273, opened 2018-05-19
Trimed values

Korbeil:
Changing value behavior by not trimming explicit values.

⚠️ Since it's changing behavior for spaces in variables, it should be added only in major version to avoid having unexpected issues.

# Related issue
- #234

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -252,7 +252,7 @@ class Loader
             }
         }
 
-        return array($name, trim($value));
+        return array($name, $value);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


vlucas:
I actually agree that explicit values in quotes should not be trimmed. I also (strongly) agree about the major version bump being required for this.

What about unquoted values? Should those be trimmed?

Korbeil:
Quick tests:

```bash
○ → export FOO=bar 
○ → echo "$FOO baz"
bar baz
```
Only one space here, the space I put after "bar" is trimmed.

```bash
○ → export FOO= bar
○ → echo "$FOO baz"
 baz
```
Space after the `=` means no variable for linux.

I would say: follow most of linux rules. Trim start & end of the string. And for variables with spaces at the start of the variable, trim theses spaces until we have a # (comment) or a variable.

PR #275, opened 2018-06-25
Fix PREG_JIT_STACKLIMIT_ERROR

danielstokes:
without the asterisk it uses too much memory, silently ignoring anything that exceeds the limits.


GrahamCampbell:
Thanks

PR #277, opened 2018-07-02
276 variable values starting with are being converted into empty strings

### LINKED ISSUE #276: Variable values starting with # are being converted into empty strings
All our slack channels had to urgently go between double quotes because of this commit https://github.com/vlucas/phpdotenv/commit/ccefc518de84decb0709d98bf2677e37b176c27d

TECH_CHANNEL=#tech to TECH_CHANNEL="#tech"

Korbeil:
Fixes #276

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -253,14 +253,14 @@ class Loader
             $parts = explode(' #', $value, 2);
             $value = trim($parts[0]);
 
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
NB The behaviour here has been changed in 3.x so that unquoted variables starting with a hash are now considered to be empty, as they will be parsed as comments.

PR #280, opened 2018-07-05
Added some additional exceptions

danielstokes:
This doesn't fix "\n" being parsed or even empty values.  It throws exceptions when preg_replace fails or does nothing.  I'm not sure if you'd like additional exception types added.  Just provide some feedback if you'd like some changes.

### DIFF
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -252,12 +252,12 @@ class Loader
                 $l = get_defined_constants(true);
                 $l = $l['pcre'];
                 $l = array_flip($l);
-                throw new \RuntimeException($l[$e]);
+                throw new InvalidFileException($l[$e]);
             }
 
-            //if they are identical, it failed to replace/match
+            // if they are identical, it failed to replace/match
             if($originalValue === $value) {
-                throw new \RuntimeException("Invalid value");
+                throw new InvalidFileException("Invalid value");
             }
             $value = str_replace("\\$quote", $quote, $value);
             $value = str_replace('\\\\', '\\', $value);
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -236,7 +236,7 @@ class Loader
                 %1$s           # match a quote at the start of the value
                 (              # capturing sub-pattern used
                  (?:           # we do not need to capture this
-                  [^%1$s\\\\]* # any character other than a quote or backslash
+                  [^%1$s\\\\]+ # any character other than a quote or backslash
                   |\\\\\\\\    # or two backslashes together
                   |\\\\%1$s    # or an escaped quote e.g \"
                  )*            # as many characters that match the previous rules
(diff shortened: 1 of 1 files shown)


src/Loader.php
```diff
@@ -236,7 +236,7 @@ protected function sanitiseVariableValue($name, $value)
                 %1$s           # match a quote at the start of the value
                 (              # capturing sub-pattern used
                  (?:           # we do not need to capture this
-                  [^%1$s\\\\]* # any character other than a quote or backslash
+                  [^%1$s\\\\]+ # any character other than a quote or backslash
```
GrahamCampbell: Why does this need to change please? Why can't we have the empty string?

src/Loader.php
```diff
@@ -246,7 +246,19 @@ protected function sanitiseVariableValue($name, $value)
...
+            if(($e = preg_last_error()) !== PREG_NO_ERROR) {
+                $l = get_defined_constants(true);
+                $l = $l['pcre'];
+                $l = array_flip($l);
+                throw new InvalidFileException($l[$e]);
+            }
+
+            // if they are identical, it failed to replace/match
+            if($originalValue === $value) {
+                throw new InvalidFileException("Invalid value");
```
GrahamCampbell: I don't think this exception message is very clear?

src/Loader.php
```diff
@@ -246,7 +246,19 @@ protected function sanitiseVariableValue($name, $value)
                 /mx',
                 $quote
             );
+            $originalValue = $value;
             $value = preg_replace($regexPattern, '$1', $value);
+            if(($e = preg_last_error()) !== PREG_NO_ERROR) {
+                $l = get_defined_constants(true);
+                $l = $l['pcre'];
+                $l = array_flip($l);
+                throw new InvalidFileException($l[$e]);
```
GrahamCampbell: I don't think the original error is going to make sense. Can we prepend some other message on the front?

src/Loader.php
danielstokes: This doesn't need to change, just slightly better regex.  This forces a minimum of 1 characters for tokens not in (\ or (double)quote).  Line 242 ")*" is that actual matching token limit within the quoted value.

src/Loader.php
danielstokes: I agree, I'll fix this.

src/Loader.php
danielstokes: I think we could, I'll think of something.

GrahamCampbell:
Ping @danielstokes. Are you able to finish this off please?

danielstokes:
yeah, I can.  I'll finish it this week.

src/Loader.php
```diff
@@ -246,7 +246,19 @@ protected function sanitiseVariableValue($name, $value)
                 /mx',
                 $quote
             );
+            $originalValue = $value;
             $value = preg_replace($regexPattern, '$1', $value);
+            if(($e = preg_last_error()) !== PREG_NO_ERROR) {
+                $l = get_defined_constants(true);
+                $l = $l['pcre'];
+                $l = array_flip($l);
```
GrahamCampbell: This won't work on PHP 7.3.

GrahamCampbell:
I've been unable to come up with a test case that actually enters any of these cases. When can this possibly fail?

PR #288, opened 2018-10-02
Allows to pass the constructer an array of file names

alar77:
The first file found is used
Filenames must be absolute , we want deterministic resolution


GrahamCampbell:
Thanks for the suggestion, but I'd rather the caller decided what file to use, than have this class do it.

PR #297, opened 2019-01-02
Support defining required env variables via file

purplekrayons:
For context, my application has a large number of environment variables that are required.
This has left me with a rather large array which makes my application entry visually unpleasing and rather unwieldy.

When attempting to define my environment variable requirements through a separate file, dotenv failed and threw this error:

> PHP Fatal error:  Uncaught Dotenv\\Exception\\ValidationException:
> One or more environment variables failed assertions: /var/www/api/init/requires.php is empty.
in /var/www/api/vendor/vlucas/phpdotenv/src/Validator.php:143

Upon reviewing the `required()` function, I found that it assumes an argument that is an array, and thus fails.

I would like to propose support for defining environment variable requirements through a separate configuration file, which returns the array. IE:

```php
<?php
# requires.php

return [
    "MY_VARIABLE_1",
    "MY_VARIABLE_2",
    "MY_VARIABLE_3"
];
```

then including it as such:

```php
# load environment variables
$dotenv = new Dotenv\Dotenv(__DIR__);
$dotenv->load();
$dotenv->required(__DIR__ . '/init/requires.php')->notEmpty();
```

This can be accomplished by first checking if the passed argument is a string, and then by checking if the passed string exists as a file. If it does, then assign the return to the variable before returning the `Validator` object.

```php
/**
* Determine if the required argument is a configuration
* file rather than a direct array
*
* @param string $variable
* 
* @return bool
*/
protected function isConfig($variable)
{
    if (is_string($variable)) {
        if (file_exists($variable)) {
            return true;
        }
    }
}


/**
* Required ensures that the specified variables exist, and returns a new validator object.
*
* @param string|string[] $variable
*
* @return \Dotenv\Validator
*/
public function required($variable)
{
    if ($this->isConfig($variable)) {
        $variable = include $variable;
    }
    return new Validator((array) $variable, $this->loader);
}
```

### DIFF
--- a/src/Dotenv.php
+++ b/src/Dotenv.php
@@ -107,6 +107,23 @@ class Dotenv
     }
 
     /**
+     * Determine if the required argument is a configuration
+     * file rather than a direct array
+     *
+     * @param string $variable
+     * 
+     * @return bool
+     */
+    protected function isConfig($variable)
+    {
+        if (is_string($variable)) {
+            if (file_exists($variable)) {
+                return true;
+            }
+        }
+    }
+
+    /**
      * Required ensures that the specified variables exist, and returns a new validator object.
      *
      * @param string|string[] $variable
@@ -115,6 +132,9 @@ class Dotenv
      */
     public function required($variable)
     {
+        if ($this->isConfig($variable)) {
+            $variable = include $variable;
+        }
         return new Validator((array) $variable, $this->loader);
     }
 


GrahamCampbell:
Just write this surely?

```php
$dotenv->required(require __DIR__ . '/init/requires.php')->notEmpty();
```

GrahamCampbell:
I'd rather separate concerns, and have the caller require the file, like in the above.

purplekrayons:
I admit, I hadn't considered that option. Though I respectfully disagree with that approach because I would argue that it is not common practice in PHP to declare includes or requires inside of functions in this way. Moreover, it feels inconsistent when looking at code; IE:

```php
$dotenv = new Dotenv\Dotenv(__DIR__);
$dotenv->required(require __DIR__ . '/init/requires.php')->notEmpty();
```

Would you consider updating the documentation instead, as I feel it's important to establish the above approach as an available option. Cheers mate :)

kyrpas:
@GrahamCampbell would it be possible to extend the parser (in a non-breaking backwards compatible way) to support indication of required vars? 

For example:
APP_URL.REQUIRED=https://www.example.com

Or
APP_URL[REQUIRED]=https://www.example.com

Or 
REQUIRED__APP_URL=https://www.example.com

Or even
*APP_URL=https://www.example.com

The indicator can be something abstract enough so that it doesn't affect anyone and also allows anyone else to use it.

What do you think?

purplekrayons:
I think the point Mr. Graham is that it's not common practice to write PHP like you've suggested, thank you for that by the way :) I would argue most people simply won't think to do that. I believe the best action here would be to note in the readme that you include external requires in the manner you suggested to me. I'm out right now, but I'll submit a PR when I get home if you'd be so kind as to include it. Thanks mate!

kyrpas:
@GrahamCampbell, you are right. My intention was to keep things DRY and avoid maintaining yet another file for indicating required variables. My solution didn't make much sense actually :) The main pain point is forgetting to include variables in the .env file. Having a separate file for the required doesn't solve this either. It's just another file that you may forget to update. 

Ideally, scanning the actual code for ENV variable references, checking them against the .env file and then complaining if any of them were missing from the .env file would solve the problem but I understand that such a mechanism is out of the scope or concern of this package as it would need to make a lot of assumptions (e.g. framework used, method of referencing env values in the code etc).

PR #299, opened 2019-01-02
Allow phpunit 6 too

GrahamCampbell:


Korbeil:
You should add low-deps & high-deps test runs in travis with that change ;) (To check if old versions are still ok)

GrahamCampbell:
No point, since we have no actual dependencies, only dev-dependencies.

PR #300, opened 2019-01-02
Refactored to use environment variables object

GrahamCampbell:
Closes #208.


GrahamCampbell:
Some examples.

---

### Example 1

The following example assumes `$path` is the full path of the directory containing your `.env` file.

This example will load the `.env` file, returning the variables, AND will have the side effect of mutating your environment.

```php
<?php

use Dotenv\Dotenv;

$dotenv = Dotenv::create($path);

$variables = $dotenv->load();
```

---

### Example 2

The following example assumes `$path` is the full path of the directory containing your `.env` file.

Similar to the first example, only this time we avoid calling any non-thread-safe functions.

```php
<?php

use Dotenv\Environment\Adapter\EnvConstAdapter;
use Dotenv\Environment\Adapter\ServerConstAdapter;
use Dotenv\Environment\DotenvFactory;
use Dotenv\Dotenv;

$factory = new DotenvFactory([new EnvConstAdapter(), new ServerConstAdapter()]);

Dotenv::create($path, null, $factory)->load();
```

---

### EXAMPLE 3

The following example assumes `$path` is the full path of the directory containing your `.env` file.

This example will load the `.env` file, returning the variables, BUT will not mutate your current environment. This is because, in example 1, the default factory was used, so the adapters that mess with your actual environment are used.

```php
<?php

use Dotenv\Environment\Adapter\ArrayAdapter;
use Dotenv\Environment\DotenvFactory;
use Dotenv\Dotenv;

$dotenv = Dotenv::create($path, null, new DotenvFactory([new ArrayAdapter()]));

$variables = $dotenv->load();
```

---

### EXAMPLE 4

The following example assumes `$content` contains the actual content of your env file:

Once again, this example will not mutate your actual environment, allowing you to inspect the contents of an env file in isolation.

```php
<?php

use Dotenv\Environment\Adapter\ArrayAdapter;
use Dotenv\Environment\DotenvFactory;
use Dotenv\Loader;

$loader = new Loader([], new DotenvFactory([new ArrayAdapter()]));

$variables = $loader->loadDirect($content));
```

PR #308, opened 2019-01-03
Fail parser on preg error

GrahamCampbell:
Replaces https://github.com/vlucas/phpdotenv/pull/280.

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -3,6 +3,7 @@
 namespace Dotenv;
 
 use Dotenv\Exception\InvalidFileException;
+use Dotenv\Regex\Regex;
 
 class Parser
 {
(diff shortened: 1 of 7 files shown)


tests/fixtures/env-wrong/bad-quotes.env
```diff
@@ -0,0 +1 @@
+FOO=\"\\\\\\\\\n
```
GrahamCampbell: @danielstokes Could you come up with a test case that causes a preg error please, so we can merge this?

GrahamCampbell:
@danielstokes Could you come up with a test case that causes a preg error please, so we can merge this?

GrahamCampbell:
Skipping locating the correct error message on super old PHP, and HHVM. I think this is good enough, since we still indicate the error was in the regular expression matching for quote.

PR #313, opened 2019-01-04
Parser error with test

danielstokes:
replaces #308


GrahamCampbell:
Thanks for providing that. Very helpful! I've incorporated your test case into https://github.com/vlucas/phpdotenv/pull/308. :)

PR #322, opened 2019-01-26
Rewrite parser and fix escaped quotes in multiline values

### LINKED ISSUE #319: Equals Symbol Causes "Failed to parse dotenv file due to an invalid name".
How do you escape the = symbol? It's causing the "Failed to parse dotenv file due to an invalid name" error because of it. It's part of a multiline value.

    SERVICE_ACCOUNT_JSON = '{
        "type": "service_account",
        "project_id": "project id",
        "private_key": "-----BEGIN PRIVATE KEY-----\nMIIE ..... DgQw==\n-----END PRIVATE KEY-----\n",
        "client_email": "email@account.com"
        "client_id": "1234567890123456789",
        .....
        .....
    }',

It is failing at the 2 == symbols at the end of the "private_key" value. If I remove the 2 symbols, it works, but unfortunately, it won't authenticate.

GrahamCampbell:
Closes #319.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,8 @@
     ],
     "require": {
         "php": "^5.4 || ^7.0",
-        "phpoption/phpoption": "^1.5"
+        "phpoption/phpoption": "^1.5",
+        "symfony/polyfill-ctype": "^1.9"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.8.35 || ^5.0 || ^6.0"
--- a/src/Lines.php
+++ b/src/Lines.php
@@ -22,7 +22,7 @@ class Lines
         foreach ($lines as $line) {
             list($multiline, $line, $multilineBuffer) = self::multilineProcess($multiline, $line, $multilineBuffer);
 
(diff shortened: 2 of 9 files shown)


drzraf:
This commit added `symfony/polyfill-ctype` to `composer.json` instead `php[ctype]`.
I can't see the rational of this choice. (And I'd rather before off this extra-dependency)

GrahamCampbell:
This adds the dependency because we are using the ctype functions.

GrahamCampbell:
The `symfony/polyfill-ctype` package has `ext-ctype` as a recommended package. When you install phpdotenv and you don't have the php extension, composer will actually warn you about this, and recommend you install it.

drzraf:
... but if one has `php[ctype]` it installs an unneeded dependency.

GrahamCampbell:
You can stop this happening in your app with the following trick. In your app's composer.json, do this:

```json
{
    ...
    "replace": {
        "symfony/polyfill-ctype": "*"
    },
    "require": {
        ...
        "ext-ctype": "*",
        ...
        "vlucas/phpdotenv": "^3.3"
    },
    ...
}
```

PR #325, opened 2019-01-28
[2.6] Backport parser fixes from 3.3.0

### LINKED ISSUE #324: variable with quote length more than 2730, getenv() return null
**v2.4.0**
I have a long json string then I put it in the`.env` file. But when I get it by the `getenv()` or `$_ENV`, it return empty.
So I test then found:
when the variable length more chan 2730,
`Loader.php -> sanitiseVariableValue() -> preg_replace()` return null.
my test .env:
`
test='111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111'
`

GrahamCampbell:
Note that this DOES NOT implement identical parsing to in 3.x. Every attempt has been made to ensure this backport behaves exactly as in 2.5.x, and also allows cases that previously crashed in 2.5.x, being as lenient as possible (where identical case in 3.3.x will hard fail).

This fixes the example from https://github.com/vlucas/phpdotenv/pull/332.

---

How does this differ from the parsing in 3.3+:

1. Unquoted values without spaces starting with a hash are considered to be comments in 3.3+ but in 2.x are treated as literal values.
2. Invalid escape sequences within quoted values in 2.5 fail silently, in 2.6 are permitted, and in 3.x will fail loudly (with a proper error message in 3.3+).

---

### Example 1:

```
FOO=#foo
```

2.5: parses as `#foo`
2.6: parses as `#foo`
3.2: parses as `#foo` (considered a bug in the 3.x series)
3.3: parses as the empty string

### Example 2:

```
FOO=#foo bar
```

2.5: parses as the empty string
2.6: parses as the empty string
3.2: parses as the empty string
3.3: parses as the empty string

### Example 3:

```
FOO="#foo"
```

2.5: parses as `#foo`
2.6: parses as `#foo`
3.2: parses as `#foo`
3.3: parses as `#foo`

### Example 4:

```
FOO="iiiiviiiixiiiiviiii\n"
```

2.5: fails silently
2.6: parses as `iiiiviiiixiiiiviiii\n`
3.2: fails with an obscure preg error
3.3: fails saying invalid escape sequence

### Example 5:

```
FOO="iiiiviiiixiiiiviiii\\n"
```

2.5: parses as `iiiiviiiixiiiiviiii\n`
2.6: parses as `iiiiviiiixiiiiviiii\n`
3.2: parses as `iiiiviiiixiiiiviiii\n`
3.3: parses as `iiiiviiiixiiiiviiii\n`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,8 @@
         }
     ],
     "require": {
-        "php": ">=5.3.9"
+        "php": ">=5.3.9",
+        "symfony/polyfill-ctype": "^1.9"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.8.35 || ^5.0"
--- a/src/Loader.php
+++ b/src/Loader.php
@@ -2,7 +2,6 @@
 
 namespace Dotenv;
 
-use Dotenv\Exception\InvalidFileException;
 use Dotenv\Exception\InvalidPathException;
 
 /**
@@ -235,42 +234,7 @@ class Loader
             return array($name, $value);
         }
 
-        if ($this->beginsWithAQuote($value)) { // value starts with a quote
-            $quote = $value[0];
-            $regexPattern = sprintf(
-                '/^
-                %1$s           # match a quote at the start of the value
-                (              # capturing sub-pattern used
-                 (?:           # we do not need to capture this
-                  [^%1$s\\\\]* # any character other than a quote or backslash
-                  |\\\\\\\\    # or two backslashes together
-                  |\\\\%1$s    # or an escaped quote e.g \"
-                 )*            # as many characters that match the previous rules
-                )              # end of the capturing sub-pattern
-                %1$s           # and the closing quote
-                .*$            # and discard any string after the closing quote
-                /mx',
-                $quote
-            );
-            $value = preg_replace($regexPattern, '$1', $value);
-            $value = str_replace("\\$quote", $quote, $value);
-            $value = str_replace('\\\\', '\\', $value);
-        } else {
-            $parts = explode(' #', $value, 2);
-            $value = trim($parts[0]);
-
-            // Unquoted values cannot contain whitespace
-            if (preg_match('/\s+/', $value) > 0) {
-                // Check if value is a comment (usually triggered when empty value with comment)
-                if (preg_match('/^#/', $value) > 0) {
-                    $value = '';
-                } else {
-                    throw new InvalidFileException('Dotenv values containing spaces must be surrounded by quotes.');
-                }
-            }
-        }
-
-        return array($name, trim($value));
+        return array($name, Parser::parseValue($value));
     }
 
     /**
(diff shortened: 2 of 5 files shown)


GrahamCampbell:
Hmm, PHP 5.3 not happy. :/

mbrodala:
For some reason lines like `foo='spaces with single quotes'` now lead to an error:

```
PHP Fatal error:  Uncaught Dotenv\Exception\InvalidFileException: Dotenv values containing spaces must be surrounded by quotes. in /.../vendor/vlucas/phpdotenv/src/Parser.php:7
```

Changing the quotes from single (`'`) to double (`"`) fixes this. However, we need single quotes to escape `$` in some values.

Is this intentional?

GrahamCampbell:
Thanks for the report. It looks like we're missing some tests.

mbrodala:
Apparently single quotes are now parsed literally in general so we cannot use them for escaping `$` at all anymore:

* `foo='$bar'` (fails)
* `foo="\$bar"` (fails)
* `foo="$bar" (works, but can lead to issues with pure shell sourcing)

GrahamCampbell:
Please could you create an issue for this, and I'll fix today.

PR #355, opened 2019-03-28
Declare the error text variable in one place and utilise it

b3none:
This means that it's only declared in one place and we don't have to switch both values.


GrahamCampbell:
Thanks for the PR, but the original implementation is both more concise and faster to execute.

PR #361, opened 2019-07-14
add php 7.4snapshot

andreybolonin:


GrahamCampbell:
Thanks for this, but I've implemented this a different way, which allows us to test on the actual RC releases of PHP 7.4, rather than the nightlies, which should be more stable. I've been experiencing losts of segfaults with various nightly builds on other projects.

PR #363, opened 2019-07-31
Handle \n in quotes strings

danmichaelo:
Background: I have a `.env` file that I also share with Docker. Docker doesn't handle newlines directly, so I escape them like so:

    VARIABLE="Line1\nLine2"

I expected this to also work with phpdotenv, so I was caught by surprise when it didn't. Is there a compelling reason not to support it, or can support be added?

I'm attaching a failing test and also a patch that makes the test go green. In my implementation, I created an array of allowed characters (`['n', 'r']`), so that it can be easily expanded if more characters should also be allowed (like `\t`). If you don't like the implementation, please let me know how it could be improved, or just change it.

Note: I guess it could be argued that `\n` should not be expanded when using single quotes (the `TEST_NS` value in the `multiline.env` test). But as long as the parser currently uses the same state (QUOTED_STATE) for both single and double quotes, it seems hard to distinguish without a larger rewrite)

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -134,6 +134,8 @@ class Parser
                 case self::ESCAPE_STATE:
                     if ($char === $value[0] || $char === '\\') {
                         return [$data[0].$char, self::QUOTED_STATE];
+                    } elseif (in_array($char, ['n', 'r'], true)) {
+                        return [$data[0].stripcslashes('\\' . $char), self::QUOTED_STATE];
                     } else {
                         throw new InvalidFileException(
                             self::getErrorMessage('an unexpected escape sequence', $value)
--- a/tests/Dotenv/DotenvTest.php
+++ b/tests/Dotenv/DotenvTest.php
@@ -296,6 +296,9 @@ class DotenvTest extends TestCase
         $dotenv = Dotenv::create($this->fixturesFolder, 'multiline.env');
         $dotenv->load();
         $this->assertSame("test\n     test\"test\"\n     test", getenv('TEST'));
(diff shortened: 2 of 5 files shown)


GrahamCampbell:
Thanks for the PR. Had this been in 2.x, it would have been a breaking change, but we are now strict about slashes in 3.x, so we can introduce new escape sequences without being breaking. :+1:

I think we'd need to support either none, or all of the common sequences. We'll definely need to add `\t`. I'm also wonder if we should add `\v`, `\e` or `\f`. I'm not sure what docker supports. Is there a list somewhere. There's a list for what PHP supports here: https://www.php.net/manual/en/language.types.string.php.

GrahamCampbell:
It looks like `\e` is probably a bad idea actually. Here's the ones C supports: https://en.cppreference.com/w/c/language/escape.

I think we should add `\f`, `\t`, and `\v` to go along side `\n` and `\r`.

GrahamCampbell:
> Note: I guess it could be argued that \n should not be expanded when using single quotes (the TEST_NS value in the multiline.env test). But as long as the parser currently uses the same state (QUOTED_STATE) for both single and double quotes, it seems hard to distinguish without a larger rewrite)

@danmichaelo Does docker only expand on double quoted strings?

GrahamCampbell:
I guess it doesn't hurt to expand for both cases, since our parser will just fail otherwise.

GrahamCampbell:
Thanks, @danmichaelo. I've merged this, and it'll be included in the v3.5.0 release. 🚀

PR #370, opened 2019-09-10
Added Validator for allowedRegexvalues

treestonemedia:
- Added a function to check if an env variable's value matches a given regex.
- This was required for our project and I think is a nice addition to this amazing tool.
- Also added a test for allowedRegexValues().
- Also added a readme block.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,11 @@ One or more environment variables failed assertions: SESSION_STORE is not an
 allowed value
 ```
 
+It is also possible to define a regex that your environment variable should be.
+```php
+$dotenv->required('FOO')->allowedRegexValues('([[:lower:]]{3})');
+```
+
 ### Comments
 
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
Thanks for the PR. I wonder if we should require exactly one match, rather than at least 1? In your usage, which behaviour would be most useful?

treestonemedia:
@GrahamCampbell good point - that was an oversight on my part - updated the PR

GrahamCampbell:
Perfect. I've made some tweaks, and merged this into master (https://github.com/vlucas/phpdotenv/commit/1bdf24f065975594f6a117f0f1f6cabf1333b156). This will be included in the v3.6.0 release.

GrahamCampbell:
This will be included in the v3.6.0 release. Thanks for the contribution! 🎉

PR #374, opened 2019-10-09
Update README.md

RichardHpa:
Not sure if this is meant to be spelt wrong but I have changed line 87 from SECRET_KEY="souper_seekret_key" to SECRET_KEY="super_secret_key"


GrahamCampbell:
Yeh, I think the misspelling was intended. ;)

PR #376, opened 2019-11-02
Multifile merge handling. Dotenv::read method.

sebastiansulinski:
Added ability to handle multiple files in the situations where there's need for more than one file to be loaded and variables to be merged / overwritten in the order of the loaded paths.

Also added `Dotenv::read()` method, which reads variables without setting them up - useful in the situations where you have to overwrite certain configuration at runtime without changing the environment.

One use case for these changes is the multi tenancy application where .env files for each tenant are stored in separate directory and only contain variables that should overwrite the ones set up in the main .env file.


GrahamCampbell:
Thanks for the PR, though this is actually already supported by using the ArrayAdapter. It was specifically designed exactly for this use cases where one wants to read the file and have nested environment variables work, but one does not want to actually mutate the environment. :)

GrahamCampbell:
Simply replace `Dotenv::create($path);` in your code with:

```php
Dotenv::create($path, null, new DotenvFactory([new \Dotenv\Environment\Adapter\ArrayAdapter()]));

**Comment by sebastiansulinski** [comment]
Oh great - thanks @GrahamCampbell

**Comment by sebastiansulinski** [comment]
One thing however - I'm trying to use it with Laravel and by the look of it its `Illuminate\Support\Env` does not support `ApacheAdapter`?

**Comment by sebastiansulinski** [comment]
Laravel `Illuminate\Support\Env::getFactory()` doesn't seem to support that?

**Comment by GrahamCampbell** [comment]
@sebastiansulinski Ah, I see. I think I've implemented what you need now in #394. Please could you give some feedback as to what you think of the implementation, and if it solves you problem now?

NB I've tried to implement it as to avoid a breaking change, where by default, we won't load all files, only the first file we can find and read.

**Comment by sebastiansulinski** [comment]
Thanks @GrahamCampbell. Am I right in thinking that this branch will not work with the latest laravel version? i.e. laravel uses v3 wheres this one is already v4?

**Comment by GrahamCampbell** [comment]
Yeh. You'll need Laravel 7 to use v4.

**Comment by GrahamCampbell** [comment]
@sebastiansulinski Yeh. You'll need Laravel 7 to use v4.

**Comment by sebastiansulinski** [comment]
Cool - I'll use my fork until L7 is released and then refactor the code. Thanks.

**Comment by GrahamCampbell** [comment]
Are you able to review the code, and establish it does what you want? I don't necessarily need you to actually run it.

**Comment by sebastiansulinski** [comment]
Yes - I've looked at the changed files including tests, which seem to return what I was after - the merged result of the multiple files, however I also need the option to return contents of these files as array without setting them up - if you check my pull `Dotenv::read()` method does exactly that.

This allows me to use it in the following way:

```php
private function lines(): Collection
{
    return new Collection(
        Dotenv::create(
            $this->envPath(),
            $this->envFile,
            Env::getFactory()
        )->read()
    );
}
```

Do you think you could add the implementation of it as well?
Not entirely sure if the `ArrayAdapter` you mentioned earlier would allow me to the same result.

GrahamCampbell:
`load` no longer returns the lines of the file, for a long time. It is a key value map of the loaded variables and their values. If you need the file contents, I'd recommend reading the files yourself, without using dotenv.

GrahamCampbell:
Maybe we could extract the file reading part of the Dotenv class to a separate class, so you can access the same logic?

sebastiansulinski:
That would be great - if it could have it as a `public static` method then that would do the trick.

GrahamCampbell:
I wouldn't want to confuse the interface of the dotenv class by allowing the internals to be called, but I would consider moving to a different class, then making it public, since that would be less confusing. Which exact method from the new code you want to call?

GrahamCampbell:
v4 is a bit different to v3 and very different from v2.

sebastiansulinski:
If there was a way to have access to methods of the Loader `findAndRead` and equivalent of `processEntries` but to only return variables as array without setting them up.

GrahamCampbell:
I suppose you don't want to have to pass paths parameter. You'd ideally like something that is just "give me the content of all the env files in an array, without passing any params"? Is that right?

I think, adding something like that to the dotenv class itself could be acceptable. I'll have a think about this, and implement something shortly.

sebastiansulinski:
I need to be able to specify paths and file names as you've just implemented as it will take main .env file plus the given tenant one that will be stored under a different path, but other than that I don't need any additional params.

GrahamCampbell:
Right, but that will be done when you instantiate the phpdotenv instance?

GrahamCampbell:
Then, this method I'm thinking of will not be static.

sebastiansulinski:
Yes - that's right.

GrahamCampbell:
Ok. I've implemented something now. Please let me know what you make of it: https://github.com/vlucas/phpdotenv/pull/394.

GrahamCampbell:
In particular, note the test:

image

sebastiansulinski:
Thanks @GrahamCampbell - appreciate your help, but would you be able to refactor the `read()` method to pass this test please?

```php
public function testDotenvReadsEnvironmentVarsMultipleWithShortCircuitMode()
{
    $dotenv = Dotenv::createImmutable($this->folder, ['.env', 'example.env'], false);
    $this->assertSame(
        [
            'FOO' => 'bar',
            'BAR' => 'baz',
            'SPACED' => 'with spaces',
            'NULL' => '',
            'EG' => 'example',
        ],
        $dotenv->read()
    );
}
```

GrahamCampbell:
Oh, then you don't need `read`. You can already use the array adapter with load?

GrahamCampbell:
Indeed, the following will already pass:

```php
    $dotenv = Dotenv::createImmutable($this->folder, ['.env', 'example.env'], false);
    $this->assertSame(
        [
            'FOO' => 'bar',
            'BAR' => 'baz',
            'SPACED' => 'with spaces',
            'NULL' => '',
            'EG' => 'example',
        ],
        $dotenv->loead()
    );
```

GrahamCampbell:
If you want to avoid actually mutating your environment, then you'll need the array adapter.

```php
$adapters = [new ArrayAdapter()];

$repository = RepositoryBuilder::create()
    ->withReaders($adapters)
    ->withWriters($adapters)
    ->make();

$dotenv = Dotenv::create(
    $repository,
    $this->folder,
    ['.env', 'example.env'],
    false
);

$this->assertSame(
    [
        'FOO' => 'bar',
        'BAR' => 'baz',
        'SPACED' => 'with spaces',
        'NULL' => '',
        'EG' => 'example',
    ],
    $dotenv->load()
);
```

GrahamCampbell:
Ok, so I've removed `read` from `Dotenv`. I think the full code you want is:

```php
<?php

use Dotenv\File\Paths;
use Dotenv\File\Reader;
use Dotenv\Repository\Adapter\ArrayAdapter;
use Dotenv\Repository\RepositoryBuilder;
use Dotenv\Loader\Loader;

$adapters = [new ArrayAdapter()];

$repository = RepositoryBuilder::create()
    ->withReaders($adapters)
    ->withWriters($adapters)
    ->make();

$content = Reader::read(
    Paths::filePaths('YOUR_FOLDER_NAME_HERE', ['.env', 'example.env'])
);

$variables = (new Loader())->load($repository, $content);
```

If the content of `.env` and `example.env` are as in the tests, then, `$variables` will look like:

```php
[
    'FOO' => 'bar',
    'BAR' => 'baz',
    'SPACED' => 'with spaces',
    'NULL' => '',
    'EG' => 'example',
]
```

GrahamCampbell:
Or, indeed, this is exactly the same as the following:

```php
<?php

use Dotenv\Dotenv;
use Dotenv\Repository\Adapter\ArrayAdapter;
use Dotenv\Repository\RepositoryBuilder;

$adapters = [new ArrayAdapter()];

$repository = RepositoryBuilder::create()
    ->withReaders($adapters)
    ->withWriters($adapters)
    ->make();

$dotenv = Dotenv::create(
    $repository,
    'YOUR_FOLDER_NAME_HERE',
    ['.env', 'example.env'],
    false // disable short circuit mode
);

$variables = $dotenv->load();
```

The only difference here is that the `Dotenv` class manages computing the paths to read, and reading the files itself, rather than you doing that, then handing over to the loader yourself.

GrahamCampbell:
This code is currently available as:

```
    "require": {
        "vlucas/phpdotenv": "^4.1@dev"
    },
```

GrahamCampbell:
I'm going to give some more time for feedback before I release 4.1.0.

sebastiansulinski:
Cheers for all this @GrahamCampbell !

PR #377, opened 2019-11-04
updated Parser class to support escaping $sign

rsalimian:
Parser raises a fatal error if env value contains a dollar sign escaping like '3mypas\$word'.
added the dollar sign to escaping list at the case ESCAPE_STATE fixed the issue. 

`Fatal error: Uncaught Dotenv\Exception\InvalidFileException: Failed to parse dotenv file due to an 
unexpected escape sequence. Failed at ['3mypas\$word'].
`

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -134,7 +134,7 @@ class Parser
                 case self::ESCAPE_STATE:
                     if ($char === $value[0] || $char === '\\') {
                         return [$data[0].$char, self::QUOTED_STATE];
-                    } elseif (in_array($char, ['f', 'n', 'r', 't', 'v'], true)) {
+                    } elseif (in_array($char, ['f', 'n', 'r', 't', 'v', '$'], true)) {
                         return [$data[0].stripcslashes('\\' . $char), self::QUOTED_STATE];
                     } else {
                         throw new InvalidFileException(


GrahamCampbell:
Does it not already parse properly without the slash?

GrahamCampbell:
> added the dollar sign to escaping list at the case ESCAPE_STATE fixed the issue.

That addition looks incorrect, and won't stop it being treated like a variable, actually.

rsalimian:
with single quotes should not treat as a variable and just as plain text.

it parses when you set like myVAR=3mypas$word, but raises that error when explicitly
set like myVAR='3mypas\$word'.

GrahamCampbell:
> with single quotes should not treat as a variable and just as plain text.

No, the variable replacement system doesn't care about the quotes you use. It is not like PHP strings.

GrahamCampbell:
Resolving of nested variables is done after parsing, so adding an escape sequence, even in principle, won't work.

GrahamCampbell:
Simply never try to escape dollars, and you will always get the string literal you want. I'm going to open a new issue to address if we want to forbid variable evaluation in single quoted strings and/or add a way to prevent variable evaluation from happening.

GrahamCampbell:
This would be a breaking change though, so would require a new major version.

GrahamCampbell:
https://github.com/vlucas/phpdotenv/issues/378

GrahamCampbell:
NB, the way docker escape dollars is with a double dollar sign. Maybe that is the way to go?

GrahamCampbell:
@rsalimian BTW, phpdotenv only actually expands environment variables that are quoted, so you should definitely write: `'3mypas$word'` and that will work correctly, both in v3 and the upcoming v4 version (with the changes from #380).

rsalimian:
thank you for having it in V4 RFC. 
I don't have control to set it like `3mypas$word`. 

this is my scenario is, in docker stack service definition
I use the export command to save the environment to a .env.  so the Export command is doing the escape stuff.

PR #384, opened 2019-11-21
PHP 8 support

Webysther:
https://travis-ci.org/Webysther/packagist-mirror/jobs/614861418?utm_medium=notification&utm_source=github_status

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,17 +3,28 @@ language: php
 php:
   - nightly
 
-jobs:
+matrix:
   allow_failures:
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Thanks, but it's too early for this just yet. PHP 8 might still make some big breaking changes yet, that we don't know about. I'd strongly recommend that no package authors allow PHP 8 yet. Anyone wanting to test packages with PHP 8 can install them by asking composer to ignore platform dependencies with a CLI flag.

PR #389, opened 2019-12-02
Fix create method return type

sunspikes:
Use correct return type for `RepositoryBuilder::create` method


GrahamCampbell:
Great, thanks!

PR #394, opened 2019-12-10
[4.1] Support multiple file names and disabling short circuit file loading

### LINKED ISSUE #393: Add ability to load multiply _ENV files
Pls, add ability to load multiply env files, smth like this:
```
$dotenv = \Dotenv\Dotenv::create(__DIR__, ['_env', '_env.db'] );
$dotenv->load();
```

P.S. it NOT a duplicate of #205 , it is a feature request

GrahamCampbell:
It is now possible to specify multiple file names, as well as multiple directories to search. By default, we'll stop looking for files as soon as we find one, like we did when there were multiple directories. To disable this behaviour, and load all files in order, one can disable "short circuit" file loading with a new parameter.

For example:

```php
// load both .env and .env2 if found
// if only one file exists, load it
// if no files exist, crash
$dotenv = \Dotenv\Dotenv::create(__DIR__, ['.env', '.env2'], false);
$dotenv->load();
```

---

Closes #393.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "4.0-dev"
+            "dev-master": "4.1-dev"
         }
     }
 }
--- a/src/Dotenv.php
+++ b/src/Dotenv.php
@@ -107,7 +107,33 @@ class Dotenv
     }
 
     /**
-     * Load environment file in given directory.
+     * Read the environment file(s), and return their raw content only.
+     *
+     * We provide the file path as the key, and its content as the value. If
+     * short circuit mode is enabled, then the returned array with have length
+     * at most one. File paths that couldn't be read are omitted entirely.
+     *
+     * @return array<string,string>
+     */
+    public function read()
+    {
+        $output = [];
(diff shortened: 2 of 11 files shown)


GrahamCampbell:
Oh, I should also mention, if more control over the exact paths tried is needed (such as only looking for some files in some directories, and files of a different name in others), instead of using the static `create` functions, you can directly "new" up the class. Please take a look at the code for the constructor, and the code for the `create` static functions. It should be obvious how to use it, with maximal flexibility from that.

GrahamCampbell:
We now support getting the raw file contents too, before they are parsed and loaded.

KarelWintersky:
I'll test it a bit later :)

GrahamCampbell:
Pushed another change. Don't want to make `read` available on the `Dotenv` class.

GrahamCampbell:
This code is currently available as:

```
    "require": {
        "vlucas/phpdotenv": "^4.1@dev"
    },
```

GrahamCampbell:
I'm going to give some more time for feedback before I release 4.1.0.

PR #408, opened 2020-01-21
Use `static` instead of `self` for extensibility

RdeWilde:
I extend `Dotenv` using my own class, but because of the `self` reference it won't initiate my own class. Changing `self`to `static` will fix this problem.

### DIFF
--- a/src/Dotenv.php
+++ b/src/Dotenv.php
@@ -67,7 +67,7 @@ class Dotenv
             $builder = $builder->shortCircuit();
         }
 
-        return new self(new Loader(), $repository, $builder->make());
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Thanks, but this is a breaking change. It is also not intended that that class would be extended. Why do you need to expend the class?

RdeWilde:
This is only a breaking change when people actually are extending the class (and overriding methods).

I extends it because I want to add methods to it. I don't see why you wouldn't allow people to extend it, or as you said, it is not intended. But if it's not, you can close it.

GrahamCampbell:
> I extends it because I want to add methods to it. 

What did you need to add?

PR #414, opened 2020-03-02
The Boolean Variables description updated

JohnMAustin78:
The boolean also accepts "true" and "false" as boolean TRUE and FALSE


GrahamCampbell:
Thanks for this. Could you update the tests too please, to both verify this behaviour, and make sure it is kept in future versions.

JohnMAustin78:
Will do

GrahamCampbell:
Looks like there are already tests. :)

PR #432, opened 2020-04-12
Add Docker setup for fast easy development

sineverba:
Add Docker / docker-compose setup for easy development.

This is not a replacement but an add-on. Every one can contribute "normally" or with the help of Docker (and swap easily between php images).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ phpstan.neon
 phpstan.tests.neon
 phpunit.xml
 vendor
+/.idea
+docker-compose.yaml
--- a/README.md
+++ b/README.md
@@ -350,6 +350,45 @@ License
 PHP dotenv is licensed under [The BSD 3-Clause License](LICENSE).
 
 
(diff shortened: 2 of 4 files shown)


GrahamCampbell:
Thanks, but I think this is more trouble than it's worth. People could already use PHP via docker without all this. For example, in my .zprofile I have:

```
alias composer74='docker run -it -w /data -v `pwd`:/data:delegated -v /Users/graham/.composer:/root/.composer:delegated --entrypoint composer --rm registry.gitlab.com/grahamcampbell/php:7.4-base'
alias php74='docker run -it -w /data -v ${PWD}:/data:delegated --entrypoint php --rm registry.gitlab.com/grahamcampbell/php:7.4-base'
alias phpunit74='docker run -it -w /data -v ${PWD}:/data:delegated --entrypoint vendor/bin/phpunit --rm registry.gitlab.com/grahamcampbell/php:7.4-base'
```

PR #447, opened 2020-06-13
Keep README file when exporting

j0k3r:
I think keeping the README file when exporting (ie: installing the project as deps) is valuable.
People will be able to read it from their vendors in case they need information from the project (or when they are working offline for example).


GrahamCampbell:
Thanks for the PR, however having thought about this, I've decided not to merge this, because:

1. I want to keep the zip files as small as possible. Lower CO2 emissions, and all that.
2. Composer already has a mechanism for pointing people at our readme, so people can get to the docs that way.
3. Our tests are important documentation, just as much as our README, so it'd only be right to include everything, or nothing.

PR #448, opened 2020-06-16
Add Repository PrefixReader and PrefixWriter

Horat1us:


Horat1us:
Can it be merged? Our product for financial company really have case to use such code.
Or should I move it to own separate package?

GrahamCampbell:
Sorry for the delay here. I was thinking about how best this should be implemented. You certainly don't need to publish a new package to get this functionality. I've come to the conclusion that the best approach is for you to create a PrefixRepository in your app which takes a repository instance and a prefix in its constructor and proxies the calls. That way we don't need to add any extra code to the core library, and you don't need a fork of the library to get what you want.

GrahamCampbell:
```php
$repo = ... // build your repo in the usual way using our factory
$repo = new PrefixRepository($repo, 'your-prefix-'); // wrap it in a class you can write an put in your app
// continue as normal
```

PR #449, opened 2020-07-01
Using non-static createArrayDotenv method call

peter279k:
# Changed log
- Using non-static `createArrayDotenv` method call because this method is not static.


GrahamCampbell:
Thanks for the PR. The mistake was that the method was not static. :)

GrahamCampbell:
https://github.com/vlucas/phpdotenv/commit/c2da5440047bbaf8ed8dc29a0df300e88b5cb50b

PR #456, opened 2020-08-26
return current instance instead of new instance

yolong-lin:
I think it will reduce the memory used maybe.


GrahamCampbell:
If you are finding that your objects are not being garbage collected, check that they are not in the global scope. If they are, you will need to `unset` them so PHP knows it can gc them.

PR #459, opened 2020-09-23
Cast for variable $string

selcukcukur:
Avoiding random selection.


GrahamCampbell:
The type of that variable is already necessarily `string`?

selcukcukur:
@GrahamCampbell Yes, no other value is returned.

GrahamCampbell:
So this PR is unnecessary?

selcukcukur:
May be required in multiple uses. @GrahamCampbell

PR #460, opened 2020-09-23
Return value does not match

selcukcukur:
The return value does not provide as stated.


GrahamCampbell:
Do you have an example where the return value is not the one that is documented?

selcukcukur:
I did not make any observations in the test environment. Just a warning I came across in the PHP Storm PHPDoc extension. @GrahamCampbell

GrahamCampbell:
This is a false positive from phpstorm.

Issue #84, opened 2015-05-05
Do not throw and exception if .env file is missing


crisu83:
I think it would be a good idea to change the `Dotenv::load` method so that it wouldn't throw an exception. The motivation for this is that not all environments use a `.env` file, for example when we deploy to Amazon we configure the environmental variables through OpsWorks. What this means in practice is that I need an additional check in application to see if the file actually exists before calling `Dotenv::load`. What do you think?

tuupola:
You could include empty .env file?

crisu83:
I could, but it's not an optimal solution because I don't want to create or keep around an empty .env file that I include in my build. To me adding an empty .env file sounds like a hack and I would like to avoid it.

vlucas:
This behavior cannot be changed at this point. Too many people rely on this behavior for it to be changed now. Just wrap your call in a try/catch.

crisu83:
That's no better than checking if the file actually exists. If the project would use proper versioning (semver) you could change this for the next major release without worrying to break BC.

KIVagant:
Please, add configuration attribute for this behaviour. I totally agree with @crisu83, this exception does not needed on production environments. Now all times I need to create empty .env file in project root.

vlucas:
You don't need to deploy an empty `.env` file - just do a `file_exists` check before using Dotenv instead.

crisu83:
@vlucas I think that's unnecessary boilerplate, if the library automatically loads a file it should ensure that the files actually exists before trying to load it.

vlucas:
@crisu83 It does, which is why it throws an Exception is it does not exist. You can also wrap the Dotenv call in a try/catch block as well.

crisu83:
Of course you can, but it feels unnecessary to add such logic to your bootstrap file. Here's how dotenv is used by Lumen, Laravel's micro framework https://github.com/laravel/lumen/blob/master/bootstrap/app.php#L5. Maybe that gives you a better idea of what I'm after here.

GrahamCampbell:
I think this behviour is correct. See how it's commented out. That's what you do if you don't want to load a .env file

crisu83:
The behavior is correct in Lumen, however if you decide to use .env files during development and run something else in production with the current implementation you will need to add unnecessary logic for this in your bootstrap file. 

[...]

PR #478, opened 2021-03-24
Adding safeLoad() information on README.md file

w3slley:
As mentioned on #84, nowhere in the README.md file contains any information regarding the safeLoad() method which allows users to suppress exceptions when no `.env` file is found.


GrahamCampbell:
Thanks. I've reverted the unrelated changes, and merged.

PR #479, opened 2021-03-28
Update wording in README.md

Krisell:
English is not my native language, so please tell me if I'm wrong here.

It looks like a word is missing in the README sentence `so it is not checked-in the code`. This PR updates this to `so it is not checked-in with the code`, i.e. a `with` is added.


GrahamCampbell:
Thanks. Both are correct, so I'm going to leave this. :)

PR #480, opened 2021-03-28
Update short circuit mode test names

Krisell:
The current test called `WithShortCircuitMode` passes `false` to disable the short-circuit, and the test called `NotShortCircuitMode` uses the default value which enables short-circuit (only loads the first found env file). This PR inverses the names of these tests, as that makes more sense in my mind.

Please let me know if I have misunderstood the meaning of these test names.


GrahamCampbell:
Thanks. I've merged this into the 4.x series.

PR #492, opened 2021-07-30
Update README.md

authanram:
Console output on copy+paste and while using the clipboard feature (appears when hovering over the code block):

Screenshot 2021-07-30 at 18 10 20

Screenshot 2021-07-30 at 18 13 20

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ dotenv](https://github.com/bkeepers/dotenv).
 Installation is super-easy via [Composer](https://getcomposer.org/):
 
 ```bash
-$ composer require vlucas/phpdotenv
+composer require vlucas/phpdotenv
 ```
 
 or add it by hand to your `composer.json` file.


GrahamCampbell:
The dollar sign is to let you know to type it into your shell. :)

authanram:
@GrahamCampbell sure, it does. Thank you for the explanation.  👍 

Maybe this should be addressed directly to github. It makes less sense to copy the shell indicator, especially if the code block is using the language identifier `shell`.

I had a short look before opening this (just to understand the thoughts i had before opening this huge pull request ^^);

Screenshot 2021-07-30 at 20 48 25

The shell indicator can't be copied here (https://symfony.com/doc/current/components/yaml.html#installation). I like that approach.

Screenshot 2021-07-30 at 20 48 47
(He's not using it)

The two big ones using opposite approaches. Obviously a matter of taste. But what is not a matter of taste is the fact, that it leads to an unexpected outcome. The user who needs this indicator to understand that this is a shell input, will fail or at least struggle because of `command not found: $` End of story. *hehe

<sub>If i'm writing to you anyway, i want to take the chance right away to say thank you for all your work: Thanks for all of your nice codes and for sharing it. :-) I used it pretty often and i loved it really much. </sub>

PR #494, opened 2021-09-17
Tipping dotenv class attirbutes

MarceloOProgramador:
In this modification I tipping the Dotenv.php attributes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ phpstan.neon
 phpstan.tests.neon
 phpunit.xml
 vendor
+.devcontainer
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
We can't do this because we need to support order PHP versions. :)

MarceloOProgramador:
But in the dotenv composer.json requirer "php": "^7.1.3 || ^8.0" and from php 7.0 is already accepted typing.

GrahamCampbell:
No, property types are only available from PHP 7.4 onwards.

MarceloOProgramador:
It's true! Sorry bro! ^^'

Issue #500, opened 2021-11-08
Add support for BOM-signed env files


hosni:
Hi!
I was working on a Laravel project that has a .env file.
But I faced an error that says:
`The environment file is invalid!`
`Failed to parse dotenv file due to an invalid name. Failed at [﻿APP_NAME].`

After some time for debugging, I realized the problem is come from encoding of the file!
The encoding of the .env file is [UTF8 BOM](https://en.wikipedia.org/wiki/Byte_order_mark)

In this [example files](https://github.com/hosni/tmp/tree/master/github.com/vlucas/phpdotenv/add-support-for-BOM-signed-env-files) you will see two files named `.env-UTF-8` and `.env-UTF-8-with-BOM`. They almost contains the same data, but `.env-UTF-8-with-BOM` starts with the UTF8 BOM, causing the parse error.

![image](https://github.com/hosni/tmp/blob/master/github.com/vlucas/phpdotenv/add-support-for-BOM-signed-env-files/.env-UTF-8.png?raw=true)

![image](https://github.com/hosni/tmp/blob/master/github.com/vlucas/phpdotenv/add-support-for-BOM-signed-env-files/.env-UTF-8-with-BOM.png?raw=true)

### How did happen?
I'm using linux as OS, but well it seems that when you add a text file in Visual Studio in windows, it adds the UTF8 BOM by default. I generated the second file that has to BOM using the Windows Explorer contextual menu "New File".

GrahamCampbell:
Yes, this is supported, but you need to tell us the encoding. https://github.com/vlucas/phpdotenv/blob/master/src/Dotenv.php#L83. :)

hosni:
> Yes, this is supported, but you need to tell us the encoding. https://github.com/vlucas/phpdotenv/blob/master/src/Dotenv.php#L83. :)

I tell the encoding and use `UTF-8`, But I got `Dotenv\Exception\InvalidFileException: Failed to parse dotenv file. Encountered an invalid name at [﻿FOO].` error!

I also tried ALL encoding that `mb_list_encodings` return, but NONE of them works!

Note that `UTF-8` is different that `UTF-8 with BOM`!

I write a simple code to test it, please see [test-utf8-with-boom.php](https://github.com/hosni/tmp/blob/master/github.com/vlucas/phpdotenv/add-support-for-BOM-signed-env-files/test-utf8-with-boom.php)

[...]

PR #501, opened 2021-11-08
Add support for BOM-signed env files

hosni:
Fix https://github.com/vlucas/phpdotenv/issues/500

### DIFF
--- a/src/Util/Str.php
+++ b/src/Util/Str.php
@@ -41,11 +41,19 @@ final class Str
                 \sprintf('Illegal character encoding [%s] specified.', $encoding)
             );
         }
-
+        $converted = $encoding === null ?
+            @\mb_convert_encoding($input, 'UTF-8') :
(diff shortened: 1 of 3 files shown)


hosni:
Some checks were not successful by GitHub actions runner, But I did not encounter this problem on the local system when run test command!

So, I run the test command more and more until I face the above problem!
I realized the problem is sometime the `uuid` function of PHP generates a UUID that is like [Scientific Notation](https://en.wikipedia.org/wiki/Scientific_notation)
For example: `61898400105e6` or `61898400105e7`
And for unknown reason, PHP treated this string as `int`! (even with strict_types)
I put a `var_dump` to print the generated key and value pair, see:
image

Actually, I don't know why string is treated as int by PHP (even with strict_types)
But I think one of the solutions is to add a prefix to `uuid` method, something like: `pdenv`, But I don't know it is a correct solution for solve this or not!

GrahamCampbell:
Thanks. 🍻

PR #527, opened 2022-07-04
Use arrow functions

MathiasReker:
Anonymous functions with one-liner return statement must use arrow functions.


GrahamCampbell:

PR #529, opened 2022-07-04
No superfluous elseif

MathiasReker:
Replaces superfluous elseif with if.


GrahamCampbell:

PR #532, opened 2022-07-06
fix(docs): correct spelling of automatically

BulegaHassan:
correct spelling of automatically in the readme from automagically


BulegaHassan:
Welcome, automagically exists and has a similar meaning

PR #558, opened 2023-08-17
Document why superglobals can appear empty

pfrenssen:
I had a case on a shared hosting setup that the `$_ENV` variable was not being set. This was due to the `variables_order` setting in `php.ini`.

Maybe it is a good idea to document this in the README?


Zubs (approved):

GrahamCampbell:
Sure, thanks.

PR #560, opened 2023-10-04
Test against php 8.3

sergiy-petrov:


GrahamCampbell:
Thanks, though this change required a little more work, and some upstream changes. This is now complete, in master, and v5.6.0 will be released shortly.

PR #572, opened 2024-07-13
Prepare for PHP 8.4

staudenmeir:
PHP 8.4 deprecates implicitly nullable parameter types:
https://wiki.php.net/rfc/deprecate-implicitly-nullable-types


GrahamCampbell:
Thanks for preparing these changes. 👍

PR #593, opened 2026-05-28
README: Prepend backslash to namespaced classes

cweiske:
Because non-backslashed classes with namespaces are relative, not absolute.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,14 +101,14 @@ SECRET_KEY="abc123"
 You can then load `.env` in your application with:
 
 ```php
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Thanks, but this is not necessary, and I feel bad taste. These examples are written outside of a namespace.

cweiske:
I copied one of the examples into my code and it did not work because the backslashes were missing.
-- 
Mit freundlichen Grüßen,
Christian Weiske

PR #598, opened 2026-08-21
Fix parsing of export with a single-character variable name

semx:
A dotenv file containing an export line with a single-character variable name fails to parse:

```
export A=1
```

```
Dotenv\Exception\InvalidFileException: Failed to parse dotenv file.
Encountered an invalid name at [export A].
```

Since the exception is thrown at file level, one such line rejects the entire file. `export AB=1` and longer names work fine.

The cause is the guard in `EntryParser::parseName()`:

```php
if (Str::len($name) > 8 && Str::substr($name, 0, 6) === 'export' && \ctype_space(Str::substr($name, 6, 1))) {
```

`export A` is exactly 8 characters, so the strict `> 8` comparison skips the prefix stripping and the raw string `export A` is then rejected as an invalid name.

This PR relaxes the guard to `>= 8` and adds regression tests for both space and tab separators. The other conditions are unchanged, so inputs like `exportFOO=...` (no whitespace after the prefix) or a variable actually named `export` are unaffected.

Full test suite passes: 252 tests, 518 assertions.

### DIFF
--- a/src/Parser/EntryParser.php
+++ b/src/Parser/EntryParser.php
@@ -96,7 +96,7 @@ final class EntryParser
      */
     private static function parseName(string $name)
     {
-        if (Str::len($name) > 8 && Str::substr($name, 0, 6) === 'export' && \ctype_space(Str::substr($name, 6, 1))) {
+        if (Str::len($name) >= 8 && Str::substr($name, 0, 6) === 'export' && \ctype_space(Str::substr($name, 6, 1))) {
             $name = \ltrim(Str::substr($name, 6), " \t\n\r\v\f");
         }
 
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Thanks for the report and the fix! I went with a slightly different approach in #599, dropping the length check altogether rather than adjusting it, since the prefix and whitespace checks already cover it. That's merged now and will go out in 5.6.5. Closing this in favour of that, but much appreciated for tracking it down.

semx:
Dropping the guard entirely is cleaner indeed, the prefix and whitespace checks already imply a valid length. Thanks for the quick turnaround!

PR #617, opened 2026-08-28
fix: reject a whitespace-only value in isBoolean()

### LINKED ISSUE #611: Validator::isBoolean() accepts whitespace-only values
`Validator::isBoolean()` accepts a whitespace-only value as valid, while an empty string is correctly rejected. `filter_var()` with `FILTER_VALIDATE_BOOLEAN` trims its input before matching, so a value that is only spaces or tabs carries no boolean but still passes.

```php
// FLAG="   " in the environment
$dotenv->required('FLAG')->isBoolean();   // passes, but should not
// FLAG= (empty) correctly fails
```

Only whitespace-only values slip through; `' on '` still validates via the filter's own trimming, and a genuinely empty value already fails. The fix is to reject a value that is empty after trimming in `isBoolean()`, mirroring the existing empty-string guard.

This is a behaviour change, since whitespace-only values pass validation today and would start failing at boot. It belongs in the next major rather than a 5.x release.

raphyabak:
## Summary

Fixes #611

`filter_var()` with `FILTER_VALIDATE_BOOLEAN` trims its input before matching, so a value that is only spaces or tabs has nothing left to match and is treated as the valid boolean `false` — it is not a "failure" that `FILTER_NULL_ON_FAILURE` would catch:

```php
var_dump(filter_var('   ', FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE)); // bool(false), not null
```

`isBoolean()` only guarded against the exact empty string (`$value === ''`), so a value that is empty only after trimming slipped past that guard and was then accepted as `false` by `filter_var()`. A genuinely empty value already failed correctly, since `'' === ''` is true — only whitespace-only values slipped through.

```php
// FLAG="   " in the environment
$dotenv->required('FLAG')->isBoolean();   // passed, but should not
// FLAG= (empty) already correctly failed
```

## Fix

Trim before the empty check, mirroring the guard `notEmpty()` already uses elsewhere in this same class, so a whitespace-only value is rejected the same way an empty one is.

As noted in the issue, this is a behaviour change — a whitespace-only value that currently passes validation would start failing — so it targets `master` rather than a `5.x` patch release.

## Test plan

- Added `INVALID_WHITESPACE="   "` to `tests/fixtures/env/booleans.env` and wired it into `invalidBooleanValuesDataProvider()`, so it's exercised by the existing `testCanInvalidateNonBooleans` / `testCanInvalidateNonBooleansIfPresent` tests alongside every other invalid-boolean case.
- Verified both new data-provider cases fail against the pre-fix code with the exact reported symptom (`isBoolean()` passing a whitespace-only value), and pass with the fix.
- Ran the full test suite (`vendor/bin/phpunit`): 282 tests, all passing (280 pre-existing + 2 new).
- `vendor/bin/phpstan analyze` (configured scope: `src`, level `max`): no errors.


GrahamCampbell:
This is a major breaking change, which is why I opened an issue to track this, instead of implementing it. Please do not just point AI an issue without maintainer approval - if I wanted to do that, I could do that myself.

